PR #5, opened 2026-05-22
WB-186: fix sse N+1 Pattern

kacpercierzniewski:
refactor(backend): wire N+1 SSE fix into the actually-used endpoint

The first pass at WB-186 introduced fetchEventsAfter + drainEventsSince but only wired them into a new GET /:id handler. That endpoint isn't called by the frontend — use-backend-execution.ts opens EventSource on the streamUrl returned by POST /execute, which is /:id/stream. Result: drainEventsSince was unit-tested and dead-pathed, while the hot SSE endpoint kept the original full-history-fetch-then-filter pattern. The N+1 bug was unchanged in production traffic. Helper-only tests masked the drift because the route had no integration coverage.

This commit:

- Restores GET /:id as a JSON metadata endpoint (REST convention, reverses the accidental co-opting into SSE).
- Rewrites GET /:id/stream to use drainEventsSince + fetchEventsAfter for both the snapshot and live drains — one query shape across the route. Heartbeat + abort cleanup preserved.
- Drops the now-unused executionEvents Drizzle import the in-route query previously needed; the route reaches the table only through fetchEventsAfter.

Cleanups:

- drain-events.ts: EventWriter and DrainResult demoted from export type to module-local types. knip flagged them as unused exports; consumer code reaches them only by inference from drainEventsSince's signature.
- drain-events.test.ts: extract an inline fixedFetcher helper to satisfy unicorn/consistent-function-scoping.

The earlier deletion of fetch-events-after.test.ts is preserved — the helper now carries zero testing scaffolding in its signature; SQL coverage moves to a separate integration test path when wanted.


apps/backend/src/routes/executions.ts
```diff
@@ -71,29 +68,17 @@ executionsRoutes.get('/:id/stream', async (c) => {
 
     let done = false;
 
+    const writeEvent = async (event: ExecutionEventRow) => {
```
piotrblaszczyk: As discussed, please also fix the pre-existing bug behind this race condition.

piotrblaszczyk (approved):

PR #7, opened 2026-05-22
WB-184: auth port

kacpercierzniewski:


apps/backend/src/routes/workflows.ts
piotrblaszczyk, line 13: Lint fails on this branch. 

ESLint flags with 'logger' is assigned a value but never used. 

This is also a silent observability regression. Without execute requested there is no way to correlate a client-side execute request with the backend-side executionId in logs. Without snapshot invalid an invalid published snapshot in production disappears from dashboards with no trace.

Proposed fix: restore both log calls. `(logger.info('execute requested', { workflowId, executionId: execution.id, sourceVersion: body.sourceVersion })` after the executions insert, `logger.warn('snapshot invalid', { workflowId, sourceVersion: body.sourceVersion, error: { issues: formatValidationDetails(snapshotParsed.error) } })` before the `invalid_snapshot return)`. This resolves the lint error and the observability regression in one change.

kacpercierzniewski: Fix applied as suggested

piotrblaszczyk (approved):

PR #28, opened 2026-06-03
feat(sdk): edge template prop

kacpercierzniewski:


packages/sdk/src/features/diagram/diagram.tsx
```diff
@@ -134,7 +139,11 @@ function DiagramContainerComponent({ edgeTypes = {} }: DiagramContainerProps) {
     [onSelectionChange],
   );
 
-  const diagramEdgeTypes = useMemo(() => ({ labelEdge: LabelEdge, ...edgeTypes }), [edgeTypes]);
+  // `useEdgeTypes` resolves the built-in default plus any Root-level
+  // `edgeTemplates`. The local `edgeTypes` prop (direct DiagramContainer mount)
+  // is merged last so an explicit per-mount override still wins.
+  const resolvedEdgeTypes = useEdgeTypes();
```
szymon-t-sc: `resolvedEdgeTypes` and `root-level` sound very mysterious and don't tell developers what they are.

Maybe something like this would be better?
```
// It returns the default built-in edge types in WorkflowBuilder,
// for example with ELK, or edges injected by plugins ex. avoiding nodes.
const defaultEdgeTypes = useEdgeTypes();
// EdgeTypes are custom edge types passed as a prop.
const diagramEdgeTypes = useMemo(() => ({ ...resolvedEdgeTypes, ...edgeTypes }), [resolvedEdgeTypes, edgeTypes]);
```

piotrblaszczyk (approved):

PR #64, opened 2026-07-10
feat(ai-studio): dock execution log beside the properties panel

librowski:
## What

**Execution Log placement**
- The log is no longer a centered overlay covering the canvas. It floats bottom-right, hugging the left edge of the properties panel when the panel is expanded, and the viewport edge when it is not. Position reacts to the panel via a ResizeObserver-based hook (`useRightPanelAnchor`).
- Height grows with entries from a single header up to a 430px cap (no fixed initial height).
- Auto-scroll follows the newest entry; scrolling away pauses it, returning to the bottom resumes it.
- Collapse state persists per browser session; starting a new execution reopens a collapsed log.

**Log entry interaction**
- The whole entry (header + detail box) is now one collapse/expand toggle. Previously the detail box showed a pointer cursor but ignored clicks.
- Selecting text inside an entry does not trigger the toggle, so log output stays copyable.
- Clicks on links/buttons inside the detail do not propagate to the toggle.

**Welcome reopen button**
- The floating info button shared the bottom-right corner with the new log position and already overlapped the expanded properties panel footer. It now hides while the log is visible or the panel is expanded.

## Verification

Exercised in the browser against a simulated execution stream: anchor behavior in both panel states, growth to the 430px cap, auto-scroll pause/resume, session persistence of the collapse state, auto-open on execution start, whole-entry toggling, and the text-selection guard. `pnpm lint`, `pnpm typecheck` (ai-studio), `pnpm format`, and `pnpm test` are green. The `apps/docs` typecheck failure (`astro check` / Starlight sidebar types) is pre-existing and unrelated.

### DIFF
--- a/apps/ai-studio/src/components/disclaimer/disclaimer-modal.tsx
+++ b/apps/ai-studio/src/components/disclaimer/disclaimer-modal.tsx
@@ -18,8 +18,7 @@ function hasAcknowledged(): boolean {
 
 export function DisclaimerModal() {
   const [open, setOpen] = useState(() => !hasAcknowledged());
-  // The reopen button shares the bottom-right corner with the execution log
-  // and the expanded properties panel - hide it while either is on screen.
+  // the reopen button shares the bottom-right corner with both of these
   const logVisible = useExecutionStore((s) => s.events.length > 0 || s.status !== 'idle');
   const { panelExpanded } = useRightPanelAnchor();
 
--- a/apps/ai-studio/src/components/disclaimer/disclaimer-modal.tsx
+++ b/apps/ai-studio/src/components/disclaimer/disclaimer-modal.tsx
@@ -3,6 +3,9 @@ import { useState } from 'react';
 
 import styles from './disclaimer-modal.module.css';
 
+import { useRightPanelAnchor } from '../../hooks/use-right-panel-anchor';
+import { useExecutionStore } from '../../stores/use-execution-store';
+
 const STORAGE_KEY = 'ai-studio:disclaimer-acknowledged';
 
 function hasAcknowledged(): boolean {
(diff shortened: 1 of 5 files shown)


apps/ai-studio/src/components/disclaimer/disclaimer-modal.tsx
```diff
@@ -15,6 +18,8 @@ function hasAcknowledged(): boolean {
 
 export function DisclaimerModal() {
   const [open, setOpen] = useState(() => !hasAcknowledged());
+  const logVisible = useExecutionStore((state) => state.events.length > 0 || state.status !== 'idle');
+  const { panelExpanded } = useRightPanelAnchor();
```
szymon-t-sc: It is good practice to use prefixes such as is, are, has, or can for boolean variables. I recommend following this convention in the future.

While TypeScript provides the type information, developers need to hover over a variable to see it. If these variables were named isOpen, isLogVisible, or isPanelExpanded (isLogCollapsed in the store), the naming would be more consistent. Right now, we are mixing prefixed and non-prefixed boolean variables.

Using boolean prefixes has two main benefits. First, the variable's type is immediately obvious from its name. Second, names like open or logVisible can be interpreted as verbs, which is not ideal. If something is named open, expand, or logVisible, a developer could reasonably expect it to be a function that can be called, such as open(), expand(), or logVisible().

apps/ai-studio/src/stores/use-execution-store.ts
```diff
@@ -24,30 +24,50 @@ type ExecutionStore = {
...
+  }
+}
+
+function persistLogCollapsed(logCollapsed: boolean) {
+  try {
+    sessionStorage.setItem(LOG_COLLAPSED_STORAGE_KEY, String(logCollapsed));
+  } catch {
+    // storage unavailable
+  }
+}
```
szymon-t-sc: Zustand persist has this feature build in.

https://zustand.docs.pmnd.rs/reference/integrations/persisting-store-data

>Or you could allow only specific fields using the following:
```
export const useBoundStore = create(
  persist(
    (set, get) => ({
      foo: 0,
      bar: 1,
    }),
    {
      // ...
      partialize: (state) => ({ foo: state.foo }),
    },
  ),
)
```

szymon-t-sc: With that we can do below:
```
useExecutionStore.setState((state) => { ...emptyStore, logCollapsed: state.logCollapsed });
```

szymon-t-sc (approved):

lukasz-jazwa (approved):

PR #68, opened 2026-07-19
feat(i18n): add Simplified Chinese (zh) translation — based on v2.2.0

welleelin:
## Summary
Added complete Simplified Chinese translation for Workflow Builder, tested and working on v2.2.0 release.

## Changes
- **zh-cn.ts**: 213 translation keys covering all UI sections
- **i18n/index.ts**: Registered zh locale, removed load:languageOnly (caused locale truncation), changed fallbackLng to zh
- **demo/index.html**: Synchronous localStorage script before ES module import for default zh

## Technical Notes
- Resource key uses `zh` (not `zh-cn`) because `load: 'languageOnly'` strips region codes
- Currently based on v2.2.0 (not main) due to main branch i18n regression
- All 213 keys tested: 文件夹名称 / 节点库 / 属性 / 关闭 / 智能体 / 删除 / 保存

## Screenshots
After fix: http://localhost:4321 shows Chinese by default:
```
文件夹名称 / demo  EN  节点库  属性  React Flow
```

## Note
Main branch has an i18n regression where translations show raw keys (header.folderName instead of translated text). v2.2.0 works correctly. This PR is based on v2.2.0 to ensure translations function properly.


piotrblaszczyk:
Hi @welleelin, thank you for this contribution! It's the first external PR to Workflow Builder, which makes it a small milestone for us.

We reviewed the translation itself and it's in great shape. All 153 keys from en.ts are covered and the quality is solid.

Honest status: we're heads-down finishing our current scope of work. We also have an i18n rework planned, and it will shape how languages beyond en/pl are exposed, so we don't want to rush that decision inside this PR. We'll come back here once the current stretch wraps up. The PR can stay open.

Thanks again for the effort and for testing it end to end!

PR #70, opened 2026-07-23
fix(sdk): TextArea is highlighted; empty required strings trigger an error

szymon-t-sc:


packages/sdk/src/features/json-form/controls/text-area-control/text-area-control.tsx
```diff
@@ -35,6 +35,7 @@ function TextAreaControl(props: TextAreaControlProps) {
         onChange={onChange}
         onBlur={onBlur}
         size="medium"
+        error={errors.length > 0}
```
szymon-t-sc, line 38: **Before:**
Image

**After:**
Image

packages/sdk/src/features/json-form/controls/text-control/text-control.tsx
```diff
@@ -32,10 +32,7 @@ function TextControl(props: TextControlProps) {
   function onBlur() {
     const trimmed = inputValue.trim();
 
-    if (trimmed === '') {
-      // eslint-disable-next-line unicorn/no-useless-undefined
-      handleChange(path, undefined);
-    } else if (isNumberInput) {
```
szymon-t-sc, line 38: `''` is saved as regular string.

We set `undefined` earlier to remove that property, allowing the validator to mark the field as required. This behavior was not implemented in TextArea, where `''` was set instead, so the required validation error was not triggered.

This change unifies the behavior between the two fields.

Why use this approach instead of adding `undefined` to textarea? This is primarily a validator issue and should be fixed there. Removing the property works for new props, but if we set defaultProperties with empty `''` (which is fine) validation will not work correctly because it assigns undefined `onBlur` in the control.

Having clear default values like `''`, which match the expected type, is preferable to relying on control behavior.

PR #76, opened 2026-08-14
feat(sdk) Variable enhancement 

szymon-t-sc:
Workflow Builder introduced variable support some time ago, allowing sidebar controls to use variables from previous nodes or global variables.

This PR enhances the feature with the missing logic to make variable usage more robust and flexible.

### Variables provided by nodes to downstream nodes can now depend on:
  - **selected type in the sidebar** - for example, selecting a time-based trigger returns schedule information, while selecting an event-based trigger returns information about the event type
  - **sourceHandle** - if a node has success and error source handles, it can provide the actual node output to the success branch and the error output to the error-handling branch

https://github.com/user-attachments/assets/cb9b1c0f-14d9-4b6a-87d2-65e6deee5765
  
### **Node variables store:**

Previously, variables from previous nodes were calculated when the control was mounted. We traversed the graph and calculated their values at the same time - now we keep nodes variables in the store and only collect them:
  - The list of variables further down the flow is calculated faster, as we only collect the variables that are available
  - It's now easier to check which variables are available from a node, and we can easily add a plugin to display them
  
obraz

### New single variable control (previously used in conditions)

https://github.com/user-attachments/assets/9eff4ab4-b77a-43fb-9d5f-4d67b89192aa

It shows a date picker with `{}` when date variables are available.

## Fixes
- When an edge was added or removed, we sometimes had incorrect highlighting of values in controls. This is now fixed thanks to `totalVariables`
- Global variables cannot be added in read-only mode

It's worth noting that the majority of the changes are encapsulated in `packages/sdk/src/features/variables/`. So while reviewing, we can identify what is used externally as a useful public API and note the core functions that could be valuable to expose.


apps/demo/src/app/data/nodes/action/action.ts
```diff
@@ -14,10 +14,15 @@ export const action: PaletteItem<ActionNodeSchema> = {
...
   outputSchema: {
     type: 'default',
-    properties: {
-      status: { type: 'string', label: 'Status', description: 'Execution status: success, failure, or skipped' },
-      result: { type: 'object', label: 'Result', description: 'The data returned by the action' },
-      errorMessage: { type: 'string', label: 'Error Message', description: 'Error details if the action failed' },
+    bySourceHandle: {
+      success: {
+        status: { type: 'string', label: 'Status', description: 'Execution status: success, failure, or skipped' },
+        // TODO: outputSchema and schema properties should support the full JsonSchema7 type imported from @jsonforms/core to build suggestions not only for objects but also for their variables.
```
szymon-t-sc: While the current format of suggestions is simple, its simplicity is limiting. For example, if we used a schema format similar to the node schema, with nesting and options, we could support much richer suggestions.

Even for simple non-object properties, such as strings, the schema can define options. We could support those here as well, allowing the condition builder to show a select with predefined options instead of an empty text field when someone picks variable received from the node in condition node later.

For object properties, returning just 'object' doesn't give us much to work with. If the object includes defined properties, we can build conditions based on those nested properties.

We should replace this format with a schema and use it similarly to how we use node schemas: to build dedicated controls for conditions and for 'set a value' nodes, allowing properties to be overridden while following the schema's types.

szymon-t-sc: Done while adding the deprecated feature.

packages/sdk/src/features/variables/utils/json-schema/get-flattened-properties-from-json-schema-7.ts
```diff
@@ -0,0 +1,70 @@
...
+export function getFlattenedPropertiesFromJsonSchema7(schema: JsonSchema7): FlattenedPropertiesIndex {
+  if (schema.type === 'object') {
+    const fields = getFlattenedPropertiesFromJsonSchema7Properties({
+      properties: schema.properties,
+      namePrefix: '',
+      rootSchema: schema,
+    });
+
+    return {
+      response: {
```
piotrblaszczyk: It injects a synthetic `response` variable into every `schemaOutput`**

Every node using the new `schemaOutput` ends up advertising `{{nodes.<id>.response}}` in the suggestions index, even when its schema declares no such property (e.g. the demo trigger only outputs `eventType` / `timestamp`). That reference is a promise the execution side doesn't keep: `response` is an ai-studio convention (its worker really does store output under that key), not a WB-wide contract. A generic engine won't have it, and `resolve-template` is intentionally strict, so a user who picks "Response" from the picker gets a run that fails on a variable the editor itself suggested.

Why nothing caught it: the injected entry is typed `object`, and all stock controls filter objects out (text controls, `VariableDynamic`, conditions), so it's invisible in the demo UI. And there's currently no place in the repo that both uses `schemaOutput` and executes (demo isn't executed; ai-studio's ai-agent is back on the deprecated `outputSchema`, which doesn't inject). It will surface through the newly public `getNodeVariablesSuggestions`, the planned variables plugin, and any future object-accepting control or schema-based validation.

Question: is there a concrete intent behind this (a "reference the whole output" feature)? If yes, let's make it an explicit, documented opt-in once the runtime actually guarantees the key. If not, my recommendation is to drop the injection: a node that exposes its whole result under `response` can declare it in its own schema, like ai-studio's ai-agent does today.

Side notes if it stays: the label is hardcoded English (bypasses i18n), and a schema that genuinely declares `response` wins over the injected one via spread order, so removal is backward-safe for those nodes.

packages/sdk/src/index.ts
```diff
@@ -228,7 +230,6 @@ export {
   EDGE_OFFSET,
   SELF_CONNECTING_EDGE_LABEL_OFFSET,
 } from './features/diagram/edges/edge.consts';
-export { VARIABLE_NODES_KEY } from './features/variables/constants';
```
piotrblaszczyk, line 231: `VARIABLE_NODES_KEY` silently dropped from the public API**

This export existed in 2.3.0 and is listed in the typedoc coverage decision log, so removing it in a minor is a semver break, even if a tiny one. Either re-add it with `@deprecated`, or keep the removal and call it out explicitly in a changeset. Trimming the rest of the surface (raw store etc.) was the right call since those were never released.

.changeset/sdk-variables-json-schema.md
```diff
@@ -0,0 +1,5 @@
+---
+'@workflowbuilder/sdk': patch
```
piotrblaszczyk: **Bump says `patch`, content says feature**

Per-handle variables + JSON Schema outputs is a new capability, so this should be `minor`. No practical effect on the released version (the other changesets already force a minor), but the CHANGELOG entry will land in the wrong section.

apps/docs/src/content/docs/node-schemas/form-controls.md
piotrblaszczyk, line 1: **`VariableDynamic` is missing from the controls reference**

The control has a changeset and demo usage, but no section here. Needs the usual table: `variableType`, `placeholder`, value contract (plain value vs a single `{{...}}` reference), and a pointer to the variable picker guide.

packages/sdk/src/features/variables/hooks/use-node-variables.ts
```diff
@@ -0,0 +1,81 @@
...
+      nodes,
+      edges,
+      excludeTypes,
+      includeTypes,
+    });
+
+    // Variables can’t change while the modal containing them is in use, so we only need to refresh them when they change.
+    // .length is critical here for performance.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [nodeId, edges.length, nodes.length]);
```
piotrblaszczyk: The `nodes.length` / `edges.length` deps (and the `key={totalVariables}` remount in VariableDynamic) are a documented perf trade-off, and the sidebar remount masks most cases. Remaining gaps: rewiring an edge to a different source (same counts) and the 100 ms indexing race after `addNode`. A cheap fix later: subscribe to the suggestions-store state identity (one new object per refresh) instead of counts.

packages/sdk/src/features/variables/utils/json-schema/get-flattened-properties-from-json-schema-7.ts
szymon-t-sc: obraz

Good catch. It was wrongly inspired by an AI Studio response, and I'll remove it.

It does have some potential. Passing the output of one node directly as the payload to another node could be useful-but since we don't currently deal with objects like that in the app, we can consider adding it when needed.

packages/sdk/src/features/variables/hooks/use-node-variables.ts
szymon-t-sc: **Fix for 100ms**
Good catch with the 100ms indexing race. It didn't appear while using the app normally because we would have to connect something and then click the node within 100ms. I've added lastUpdateTimestamp, which should solve race-condition issues like this (suggestions store force refresh when any suggestion is changed) now hook dependecy array looks like this `[lastUpdateTimestamp, nodeId, excludeTypes, includeTypes, edges.length, nodes.length]);`.

**Not cheap fix**
>A cheap fix later: subscribe to the suggestions-store state identity (one new object per refresh) instead of counts.

For clarity, suggestion store caches suggestions produced **by the** node, divided by its `sourceHandles`.

In `useNodeVariables` (here), we collect those suggestions by determining the previous nodes connected to our `nodeId`. This is calculated live for the hook when `nodes.length` / `edges.length` changes.

Traversing the graph is expensive, previous implementation also did that. The difference is that it now collects cached suggestions from the store instead of building them live (earlier version) - which is good because building them for different handles and conditions is more complex now.

And since we can't clearly determine which edge changes impact a `nodeId`'s variables, watching `nodes.length` and `edges.length` makes sense.

**Why I believe my approach is the right one**
The question 'which variables are available for a given nodeId?' is, in my mind, a business-logic question. We shouldn't calculate the answer by relying on the React and Zustand hooks lifecycle, as it's less clear and can't be used outside React components.

We already have a dedicated `getAvailableVariablesByNodeId` function that can be called outside. Relying fully on hooks here would duplicate the logic.

Keeping `getAvailableVariablesByNodeId` as the basis of the implementation is better than splitting it. `getAvailableVariablesByNodeId` is used by the validation plugin and can currently be called when imported. If we only had a hook-based solution, we would need to entangle the validation function with the React component lifecycle, which is unnecessary. Keeping these functions independent like this allows us to easily adapt them to run on the server - we just need to pass the store JSON to them.

**Regarding the `totalVariables` key**
`key={totalVariables}` is mainly a result of the library we are using for the variable picker. Changing the suggestions property works, but the control behaves more reliably when we don't do it.

packages/sdk/src/index.ts
szymon-t-sc: Restored.

packages/sdk/src/features/variables/hooks/use-node-variables.ts
piotrblaszczyk: Idea is very good, but the new deps defeat the memo entirely: the ancestor BFS now runs on every render, including every drag frame.

Solution should be pretty straightforward:

```ts
const NO_TYPES: VariableType[] = [];   // module scope

const { excludeTypes = NO_TYPES, includeTypes = NO_TYPES } = options || {};
```

packages/sdk/src/features/json-form/controls/variable-dynamic-control/variable-dynamic-control.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+      handleChange(path, trimmedValue || undefined);
+    },
+    [handleChange, path, variableType],
+  );
+
+  return (
+    <ControlWrapper {...props}>
+      <DynamicTypedVariableOrInput
+        key={totalVariables}
```
piotrblaszczyk: `key={totalVariables}` now remounts a focused input: since suggestions refresh reactively, editing an ancestor while this input is focused changes the count and the remount drops focus, caret, and the open popup

'VariableText` reads suggestions from props each render, so the key looks unnecessary. Remove it here and at `variable-text-control.tsx:39` and `dependencies.tsx:32`; if the mentions lib truly needs a remount, key off content (join of suggestion ids), not a count, and never while focused.

packages/sdk/src/features/variables/stores/core/set-suggestions-node-data/set-suggestions-node-data.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export function setVariablesSuggestionsNodeData(params: Params): VariablesSuggestionsStore {
+  // Pass the store if you want to call this function multiple times
+  let storeToMutate =
+    params.shouldOnlyPassedStore === true ? params.cachedStore : useVariablesSuggestionsStore.getState();
+
+  if (params.type === SUGGESTION_NODE_TYPE.COMMON) {
+    storeToMutate = {
+      ...storeToMutate,
+      lastUpdateTimestamp: Date.now(),
```
piotrblaszczyk: `Date.now()` as change token collides: two refreshes in the same millisecond produce an equal value, zustand bails on `Object.is`, and mounted hooks miss the second write. Also here on line 65, and `refreshAllSuggestions` resets it to `0`. Please use a monotonic `version++` counter instead.

packages/sdk/src/index.ts
```diff
@@ -260,3 +261,45 @@ export { Icon } from '@workflow-builder/icons';
...
+// variables. Attach it via `NodeData.schemaOutput`. Two forms:
+// `{ type: 'default', bySourceHandle }` maps each source handle to a JSON
+// Schema (`every` covers all handles at once, e.g. `success` / `error` split);
+// `{ type: 'variant', variants }` picks the shape at runtime from a property
+// value, for nodes whose outputs depend on how they are configured.
+export type { NodeSchemaOutput } from './node/node-output-schema';
+
+// Every variable a node can reference — global variables plus the outputs of
+// its ancestors — grouped for a picker UI. Optionally narrowed by variable type
+// (`includeTypes` / `excludeTypes`). Recomputes only when node/edge counts
```
piotrblaszczyk: Stale contract: it now recomputes on every suggestions-store write by design (and, until the deps fix, on every render). Rewrite this and the comment at `use-node-variables.ts:64`, which claims variables can't change while mounted - the fix exists because they can.

packages/sdk/src/features/json-form/controls/variable-dynamic-control/variable-dynamic-control.tsx
szymon-t-sc: Changed total to key. I used total because that was enough in the current setup of the app.

>editing an ancestor while this input is focused changes the count and the remount drops focus, caret, and the open popup

We can't change ancestors while the input is focused. The only way we can change ancestors without closing the node sidebar is by drawing an edge, which loses focus by definition.

In the future, when we have multiple people working on the workflow and synchronization that changes ancestors in the background with websockets, we can think about how to keep the focus then. It's just problematic.

The videos below show an artifact/error caused by changing ancestors for a string input in simple textarea, but dynamically typed inputs have {} shown conditionally if there are variables of the selected type. If we have a variable set there and someone removes the list of ancestors, we have another list of edge cases to solve, key just rerenders them again.

I'm open to tackling that, but I would handle it when it actually happens.

>if the mentions lib truly needs a remount, key off content (join of suggestion ids), not a count, and never while focused.

One of the problems solved with a key:

Without it
https://github.com/user-attachments/assets/6606ec2f-5222-4c16-8ebe-ca0cb1c2e5b5

With it
https://github.com/user-attachments/assets/c68e03ef-4c49-4fe5-b951-c1942a15442b

PR #79, opened 2026-08-18
fix(execution-worker): Fix sequence numbering | fix(execution-core): Route node_started event through error policy

dawidaksamski:
## What

WB-429 — two bugs in execution event handling, both reachable in ordinary parallel runs.

**Sequence numbering was racy.** The worker derived `sequence` as `MAX(sequence) + 1` inside the
INSERT, against a unique index on `(execution_id, sequence)`. Concurrent inserts all read the same
maximum, so any parallel wave in the graph runner collided, the index rejected the losers, and
Temporal retried `emitEvent` until its attempt budget was gone.

The number is now minted in workflow code (`sequenced-event-emitter.ts`), where it's deterministic
under replay and uncontended — the worker is the only writer of `execution_events`. The insert takes
it as a parameter and adds `ON CONFLICT (execution_id, sequence) DO NOTHING`, which also fixes a
related bug: a retry after a lost ack used to recompute `MAX+1` and write a *second copy* of the
event under a new number.

**A failed `node_started` emit bypassed the error policy.** It was emitted above `runNode`'s try
block, so a failure there escaped the graph runner entirely — no `node_failed` event, `errorPolicy`
never consulted. Moved inside the try, so `continue` / `errorRoute` absorb it and `fail` aborts
through the normal path.

## Verification

- New tests: sequence uniqueness across a parallel wave, replay-equivalence of the numbers, and
  three error-policy cases for a failing `node_started`. Both fixes mutation-tested.
- SQL checked against a live Postgres in a rolled-back transaction: the old shape reproduces the
  exact `duplicate key value violates unique constraint "execution_events_execution_sequence_idx"`,
  the new shape inserts a full wave and silently discards a replayed `(execution_id, sequence)`.
- `pnpm test` (367) green, lint and typecheck clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,21 +6,21 @@ Visual workflow editor SDK (React) with a reference backend and Temporal-based e
 
 Three onboarding paths (A installs from npm; B, C run the repo locally). README "Get started" covers all three. Path A ("Embed the SDK") installs `@workflowbuilder/sdk` from npm; the README has install + a minimal snippet, and the full guide lives in the [docs site](https://www.workflowbuilder.io/docs/get-started/quick-start/wb-as-react-component/).
 
-| Command                      | Path | What it does                                                                            |
-| ---------------------------- | ---- | --------------------------------------------------------------------------------------- |
-| `pnpm preflight`             | B/C  | Verify Node / pnpm / Docker / ports / `.env` files. Add `--json` for agents             |
-| `pnpm dev` / `pnpm dev:demo` | B    | Demo (UI only, port 4200). No backend, no Docker                                        |
-| `pnpm infra:up`              | C    | Start Postgres + Temporal in Docker. Required before backend/worker                     |
-| `pnpm -F backend db:migrate` | C    | Apply Drizzle migrations out-of-band (backend also auto-migrates on boot)               |
-| `pnpm dev:ai-studio`         | C    | Full stack: infra + backend (3001) + worker + AI Studio frontend (4201)                 |
-| `pnpm dev:backend`           | C    | Backend only (debug). Needs infra up                                                    |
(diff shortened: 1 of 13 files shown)


dawidaksamski:
## Follow-up in this PR: serialized event emits

Assigning sequence numbers in the workflow removed an invariant the read path silently
depended on. The SSE drain advances a monotonic cursor and queries `sequence > cursor`, so
it needs rows to become **visible** in ascending sequence order. The old `MAX(sequence) + 1`
insert guaranteed that by accident: a colliding insert failed and recomputed its number only
after the winner had committed. With numbers minted up-front, a parallel wave's inserts run
on separate connections and can commit out of order — the cursor then steps past a row that
hasn't landed, and that event is never selected again. A dropped `node_completed` leaves the
node stuck at `running` in the canvas for the rest of the run.

Fixed by chaining emits so only one `emitEvent` is in flight per execution, which restores
commit order. Two promises: the caller gets one that rejects on failure (so it still reaches
`errorPolicy`), while the chain link is `write.catch(() => {})` and always fulfils — with a
bare `tail = write`, one failed emit would skip every later link and silently drop the rest
of the run's events.

Cost: emits are staggered by one activity round-trip instead of overlapping. Gaps in the
numbering remain possible (a cancelled run, or an emit that exhausts its retries) and are
harmless — the cursor only needs ascending arrival, not contiguity.

Verified with six new writer-side tests plus a `drain-order.test.ts` on the backend that
documents the cursor contract from the reading side. Both halves are mutation-tested: removing
the chain fails four tests with wave-scoped inversion (`[1, 2, 3, 6, 5, 4, 9, 8, 7, …]`),
and reverting the poison guard fails two.

piotrblaszczyk (approved):

PR #81, opened 2026-08-18
Trigger PR checks on the ds-2.0 integration branch

librowski:
Pre-step of the token-pipeline task. The DS 2.0 migration lands component work as PRs targeting a long-lived `ds-2.0` integration branch; `pr-check.yml` and `pr-check-docs.yml` trigger only on PRs to `main`, so those PRs would run with zero CI. Two-line trigger extension, no behavior change for PRs to `main`. Mergeable immediately.

### DIFF
--- a/.github/workflows/pr-check-docs.yml
+++ b/.github/workflows/pr-check-docs.yml
@@ -12,6 +12,8 @@ on:
   pull_request:
     branches:
       - main
+      # DS 2.0 integration branch — same feedback as PRs to main.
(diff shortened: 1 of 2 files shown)


librowski:
Folded into #82 — a pull_request run uses the workflow from the merge ref, so the trigger change rides in the PR itself and, once merged, sits on ds-2.0 for every later PR targeting it. A separate PR to main was unnecessary.

PR #84, opened 2026-08-20
[DS 2.0] Typography layer (wb-text-* roles)

librowski:
DS 2.0 typography layer. **Stacked on #83** (needs the token-usage lint); rebase onto `main` and retarget after #83 merges.

## Changes

- **39 `wb-text-{family}-{size}[-emphasized]` utility classes** in `packages/ui/src/styles/typography.css`, one per text style from the Figma Typography guidelines (Display, Headline, Title, Body, Label, Node, UI/Code). Regular = 400, Emphasized = 600, line-heights per style. Existing `ax-public-*` classes untouched; nothing consumes the new classes yet (first consumer: the Chips redesign).
- **Provisional `--wb-font-size-*` primitives** live in a dedicated `_provisional.css`, wired into both distribution channels (barrel import + combine globals list) so deleting the file without cleaning the wire-ups fails the build; the future unlayered DS 2.0 export overrides the layered definitions automatically, and the whole file is removed in the token-export task.
- **Poppins and Inter ship with `@workflowbuilder/ui` itself** (fonts.css on the root barrel, imports layered via `layer(ui.base)`; lib mode embeds the files as base64, as the SDK bundle always did). The SDK inherits the same 12 font faces through the JS module graph and drops its own fontsource dependencies — the classes work for standalone `ui` consumers.
- **Docs**: UI Library → Typography page (scale table, role recommendations).
- **Changesets**: minor for `@workflowbuilder/ui` and `@workflowbuilder/sdk`.

Source of truth: Figma DS page Typography (`?node-id=1-53`) — per-style table extracted programmatically; scale cross-checked against the "39 text styles" foundation note.


packages/ui/package.json
piotrblaszczyk, line 1: Knip config didn't follow the font move. `knip.config.js` still has `ignoreDependencies: ['@fontsource/poppins']` under `packages/sdk`, but this PR removes that dependency from the SDK, so the entry is now dead. Meanwhile `packages/ui` gained `@fontsource/poppins` and `@fontsource/inter` as `devDependencies` consumed only via CSS `@import` in `fonts.css`. Knip doesn't parse CSS, so `pnpm knip` will report both as unused devDependencies. Please move the `ignoreDependencies` entry from the sdk workspace to the ui workspace and add `@fontsource/inter` to it. Not CI-gating (knip isn't part of pnpm check or pr-check), so fine to do here or in a follow-up within the stack.

piotrblaszczyk:
`overview.mdx` doesn't mention fonts for the subpath path. The "Styles" section tells subpath consumers to add `@workflowbuilder/ui/styles.css` "if you also want the typography classes", but `styles.css` deliberately doesn't carry the font faces (fonts.css is only on the root barrel). Following that guidance as written, the new `wb-text-*` classes will silently render in system-ui. The typography page already documents this ("subpath-only consumers load the families themselves"); one sentence in overview.mdx pointing that out would close the gap.

piotrblaszczyk (approved):

packages/ui/package.json
librowski: Good catch — fixed in 714f7f1 (lands with this stack, in the ds2-font-assets layer where the fontsource packages now live): the ignore entry moved to the `packages/ui` workspace and covers both `@fontsource/poppins` and `@fontsource/inter`; `pnpm exec knip` is clean. The CLAUDE.md table also stopped claiming knip runs as part of `pnpm check` — it never did.

PR #87, opened 2026-08-20
[DS 2.0] Chip component

librowski:
Adds the `Chip` component and swaps the first consumer onto it.

- `Chip` in `@workflowbuilder/ui`: solid and outline treatments, sizes `s`–`xl`, optional prefix icon and close affordance, per the DS 2.0 "Chips and Tag" component set.
- The chip color tokens and `--wb-radius-50` land in `_provisional.css` until the token export ships them; the light-theme outline overlay is a `color-mix()` because `gray-100-85` does not exist in the exported palette.
- The SDK dynamic-conditions counter tag renders with `Chip`; the local `.tag` treatment and its background variable go away.
- Docs page with example and generated props table. `Chip` defines no `ax-public-*` variables (deprecated contract), so the CSS variables section renders the standard empty state.

Not swapped here: `branch-card.module.css` uses accent (`acc2`) chip colors this token factory does not cover; it migrates with the token pipeline switch.

### Component-layer checklist (template for the remaining DS 2.0 layers)

- [x] Component matches the Figma component set (variants, sizes, metrics)
- [x] Missing tokens defined in `_provisional.css`, aliased to `--ax-*` per theme
- [x] Registered in vite `componentEntries` and the root barrel
- [x] Docs: page + example + `COMPONENTS` entry in `ui-components.mjs`
- [x] SDK consumers swapped, or deferred with a stated reason
- [x] Changesets: `ui` minor (new component), `sdk` patch (visual swap)
- [x] `build:lib`, docs build, stylelint, typecheck, tests green


packages/ui/src/components/chip/chip.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export type ChipVariant = 'solid' | 'outline';
+export type ChipSize = 's' | 'm' | 'l' | 'xl';
+
+export const Chip = forwardRef<HTMLSpanElement, ChipProps>(
+  ({ label, variant = 'solid', size = 'm', prefixIcon, onClose, closeLabel, className, ...rest }, ref) => {
+    return (
+      <span ref={ref} {...rest} className={clsx(styles['chip'], styles[variant], styles[size], className)}>
+        {prefixIcon && <span className={styles['icon']}>{prefixIcon}</span>}
+        <span className="wb-text-label-s">{label}</span>
```
piotrblaszczyk, line 37: The typography page added in #84 recommends `label-l-emphasized` for "Buttons and chips", but Chip renders its label with `wb-text-label-s` (10px, weight 400) across all sizes, including xl. One of the two seems off. If the "Chips and Tag" component set in Figma really specifies Label/S Regular for every chip size, could we adjust the recommendation table in `typography.mdx` (or scope it to buttons only) so the docs don't contradict the component? And if not, should the chip label scale with size?

piotrblaszczyk (approved):

packages/ui/src/components/chip/chip.tsx
librowski: Fixed in 714f7f1 (ds2-font-assets layer of this stack): the recommendation row now names buttons alone and chips get their measured role.

packages/ui/src/components/chip/chip.tsx
szymon-t-sc, line 1: The file is named `chips/chips.tsx`. It probably should be `chip/chip.tsx` to match the rest of the application.

szymon-t-sc (approved):

PR #90, opened 2026-08-21
[DS 2.0] Rebuild the form fields against the design system

librowski:
Rebuilds `Input` and `TextArea` against the design system's field states.

**State instead of a boolean** — `state?: 'default' | 'critical' | 'success' | 'read-only'` replaces `error?: boolean`. Hover, focus and filled stay visual states rather than props. `read-only` applies the native attribute, so the value stays focusable and copyable, while `disabled` stays out of the tab order — the distinction the design guidelines call for.

**Slots** — `prefixIcon`, `suffixIcon` and an optional `onClear` affordance. Icon slots accept interactive content: clicks on buttons or links inside a slot pass through, anywhere else on the frame focuses the input.

**Sizes** — letter-scaled `l` `m` `s` `xs`. Input row heights 42/36/32/28 with padding 11/10/8/6 vertical and 12 horizontal, gap 8, radius 8/6/6/6; the field composes a label row, the input row and a helper row that wrap without painting over the control. TextArea minimum heights 120/100/80.

**Colours** — every surface, border and text colour binds to a field token. The hover border is deliberately lighter than the default one; that is what the design paints. The field composition (label and helper) exposes its own public variables, and `disabled` now dims the whole composition, muting the critical helper.

NumberField was pulled out of this PR by maintainer decision; it ships as a standalone component after the DS 2.0 migration.

Call sites migrated in this PR (`error={…}` → `state`, word sizes → letters).

Verified: ui/sdk lint + typecheck, `build:ui`, `build:lib`, stylelint, all test suites, docs build, demo build — re-run after extracting NumberField and rebasing onto the Button layer.


apps/docs/src/components/ui-examples/input.tsx
```diff
@@ -1,14 +1,48 @@
...
 export function InputExample() {
-  const [value, setValue] = useState('');
+  const [search, setSearch] = useState('');
+  const [projectName, setProjectName] = useState('');
+  const [displayName, setDisplayName] = useState('');
 
   return (
     <ComponentPreview>
-      <Input placeholder="Type something" value={value} onChange={(event) => setValue(event.target.value)} />
+      <div style={{ display: 'flex', flexDirection: 'column', gap: 'var(--wb-space-100)' }}>
```
piotrblaszczyk: why inline styles instead of css class?

apps/docs/src/components/ui-examples/text-area.tsx
```diff
@@ -4,11 +4,33 @@ import { useState } from 'react';
...
 
 export function TextAreaExample() {
-  const [value, setValue] = useState('');
+  const [description, setDescription] = useState('');
+  const [summary, setSummary] = useState('');
 
   return (
     <ComponentPreview>
-      <TextArea placeholder="Multi-line input" value={value} onChange={(event) => setValue(event.target.value)} />
+      <div style={{ display: 'flex', flexWrap: 'wrap', gap: 'var(--wb-space-100)' }}>
```
piotrblaszczyk: the same here

piotrblaszczyk (approved):

PR #92, opened 2026-08-21
[DS 2.0] Rebuild NavButton, add MenuTriggerButton

librowski:
Rebuilds `NavButton` on the same shape as the new `Button` and adds `MenuTriggerButton`.

**API** — `size`, `styleVariant` and explicit `prefixIcon`/`suffixIcon` slots. The label/icon/icon+label subtypes are gone, and with them the shared children-shape guards: nothing imports them any more, so `guards.ts` and its spec are deleted.

**Selected is not pressed** — this is the point of the layer. Selection (`isSelected`) survives hover, and pointer-down overrides it; each state draws its own token, exactly as the design defines and as the token export ships (`nav-button-bg-primary-{default,hover,pressed,active,focus,disabled}`).

**Sizes** — seven steps from the design set: boxes 48/42/36/32/28/24/20, icon-only padding 12/11/9/7/6/6/4, gap 8 (6 for the two smallest), radius 8/8/6/4/4/4/4, icon boxes 24/20/18/18/16/12/12. Nav labels keep the regular weight, unlike the emphasized button labels. `styleVariant: 'plain'` is the design's "no background" treatment — transparent by default, still lit on the other states.

**MenuTriggerButton** — new, composes NavButton and maps `isOpen` onto selection.

**SegmentPicker** keeps its public API: it consumed NavButton's subcomponent prop types, and now translates its own long-form sizes and shapes to the new slots internally.

Verified at the tip of the stack (so this covers the four layers below it): ui/sdk lint + typecheck, `build:ui`, `build:lib`, stylelint, all test suites, docs build (211 pages), demo build.


packages/ui/src/components/collapsible/collapsible.tsx
```diff
@@ -54,13 +54,13 @@ Collapsible.Button = function CollapsibleButton() {
 
   return (
     <NavButton
+      aria-label={context?.isExpanded ? 'Collapse' : 'Expand'}
```
piotrblaszczyk: I think we should use i18n instead of hardcoded labels

packages/ui/src/components/modal/modal.tsx
```diff
@@ -99,11 +99,7 @@ export const Modal = forwardRef<HTMLDivElement, ModalProps>(
                   )}
                 </div>
               </div>
-              {onClose && (
-                <NavButton onClick={onClose}>
-                  <X />
-                </NavButton>
-              )}
+              {onClose && <NavButton aria-label="Close" onClick={onClose} prefixIcon={<X />} />}
```
piotrblaszczyk: The same here. aria-label="Close"

piotrblaszczyk (approved):

packages/ui/src/components/collapsible/collapsible.tsx
librowski: These are now props with defaults, same pattern as `Chip.closeLabel`: `Modal.closeLabel` and `Collapsible.expandLabel`/`collapseLabel`.

PR #93, opened 2026-08-21
[DS 2.0] Ship the fonts as assets instead of base64

librowski:
Fonts stop being base64-inlined. Vite's library mode inlines every asset unconditionally (`assetsInlineLimit` has no effect there), so the built stylesheets carried 382 KB of font payload.

**What changed** — the `@font-face` rules are generated after Vite finishes, from the fontsource metadata, and the files are copied into `dist/assets`:

- Poppins 400 and 600 latin stay inlined — the only weights `typography.css` declares — so ordinary text needs no extra request and shows no swap flash.
- Every other face (Poppins 300/500/700, all latin-ext, Inter) is a `.woff2` fetched on demand.
- Each face now carries `unicode-range`, which the per-subset fontsource stylesheets omit. Without it a browser has to consider both subsets; with it, a document without extended latin never fetches those files.
- The legacy `.woff` source is gone — that duplication is why twelve faces produced twenty-four payloads.

**Sizes** — `ui/dist/index.css` 509 KB → 150 KB, `sdk/dist/style.css` 591 KB → 230 KB, `ui/dist/styles.css` 7 KB → 32 KB (it now carries the two inlined faces). Ten `.woff2` files ship in each package's `dist/assets`.

**Why the SDK is in scope** — it bundles the UI stylesheet into its own, so it carried the same payload; it now appends the generated block and copies the assets, reading them as artifacts from the UI build rather than importing its build code.

**New gate** — `check-built-css` fails when a stylesheet references an asset missing from `dist`. That is the failure mode this arrangement invites, and the one an earlier font change hit only on a clean CI install. Verified by deleting a file and watching the build fail.

**For consumers** — imports are unchanged, and the subpath path finally works as documented: `styles.css` now really does bring the typefaces. Two caveats, both in the changeset: a Content-Security-Policy naming `font-src` needs `'self'` (or the serving origin) rather than `data:`, and the `dist` layout has to survive copying, since the stylesheets reference `./assets/*.woff2`. Bundlers handle that themselves.

`overview.mdx` gains a short section listing what each style surface provides and what it does not — which also closes the gap a reviewer flagged on the typography PR, where subpath consumers were told to add `styles.css` for typography that then rendered in a system font.

Verified: ui/sdk lint and typecheck, `build:ui`, `build:lib`, stylelint, all test suites, docs build, demo build, and the asset gate proven by removing a font file.


packages/ui/combine-css-bundle.mts
```diff
@@ -1,34 +1,164 @@
...
+      writeCombinedStylesheet(distributionDirectory, documentationPreviewDirectory, stylesDirectory, fontStyles);
+      writeGlobalStylesheet(distributionDirectory, stylesDirectory, fontStyles);
+      appendFontStylesToEntryChunk(distributionDirectory, fontStyles);
       prependLayerOrderToAssets(distributionDirectory, stylesDirectory);
     },
   };
 }
 
+type FontFaceDefinition = {
+  family: 'Inter' | 'Poppins';
```
piotrblaszczyk, line 48: Poppins and Inter are licensed under the SIL OFL 1.1, and this PR starts redistributing their `.woff2` files as standalone assets in the npm package. The OFL asks for the license text to accompany the font files when they are redistributed. Strictly speaking this predates the PR, since base64-embedded fonts are redistribution too, but standalone files in `dist/assets` make it visible. Not a blocker for this PR. As a follow-up, let's either copy the OFL texts from the fontsource packages into `dist/assets` or add a fonts section to the package `LICENSE` file. The `check-built-css` gate could then assert the license file ships next to the assets, so it never silently drops out of the publish.

piotrblaszczyk:
Two follow-up asks, neither blocking. First, could you add a `packages/ui/font-assets.decision-log.md` for this arrangement? The PR description already carries the context, the decision, and the numbers. What is worth capturing on top are the rejected alternatives: keep everything inline, ship everything as assets, patch the Vite config, or use a CDN. This setup is exactly the kind a future refactor might "simplify" back to fontsource imports, silently reintroducing the 382 KB. 

Second, the docs give one preload example, but consumers have to inspect `dist/assets` to learn the other file names. A small table in `overview.mdx` listing the twelve shipped faces (family, weight, subset, file name, inline or asset) would close that gap.

piotrblaszczyk (approved):

packages/ui/combine-css-bundle.mts
librowski: Good catch - fixed in 6d0d615. The bundle step now copies the license texts from `@fontsource/poppins` and `@fontsource/inter` into `dist/assets/` as `OFL-poppins.txt` / `OFL-inter.txt`, with a build test asserting both exist.

librowski:
@piotrblaszczyk both done: `packages/ui/font-assets.decision-log.md` records the layout decision with the rejected alternatives (all-inline, all-assets, Vite patch, CDN), and `overview.mdx` now has the 12-face table (family / weight / subset / file / inline-vs-asset) - 6d0d615. The log is also registered in `DECISION-LOGS.md` (c082aca; the collector needed a missing `Date` header fixed on an older log, which surfaced seven unregistered entries).

PR #109, opened 2026-08-28
feat(temporal): add @workflowbuilder/temporal plugin package

piotrblaszczyk:
# feat(temporal): package the Temporal integration as a plugin (WB-526)

Turns our Temporal integration into an installable Temporal **Plugin** and moves the
worker and the backend onto it. Today the integration is app-level wiring: to run a
Workflow Builder diagram on your own Temporal you would have to copy our worker.

## Commits, in order

| Commit | What |
| --- | --- |
| `07049f2f` | Adds `packages/temporal`: the plugin, three entry points, 25 tests, release workflow, CI job |
| `d27d231c` | Repoints the worker and backend onto it, deletes what moved |
| `4f15e667` | Fixes cancel in AI Studio: the UI closed the SSE stream before the cancel event arrived |
| `e825d74b` | Aligns all four `@temporalio/*` packages on 1.23.0 |
| `ce43beef` | Updates docs that pointed at the moved files |

The first two are the substance; read them in order.

## What the worker looks like now

```ts
const plugin = new WorkflowBuilderPlugin<AiStudioNode>({
  executors: { 'ai-studio/trigger': executeTrigger, /* ... */ },
  store: database,
});

const worker = await Worker.create({
  connection,
  taskQueue: plugin.taskQueue,
  workflowsPath: fileURLToPath(new URL('workflows.ts', import.meta.url)),
  plugins: [plugin],
});
```

`store: database` needs no adapter: the worker's database module already had the two
methods the port asks for. Gone from the worker: three hand-written activities, the
local task queue constant, the `resolveExecutor` wiring, and 4 files that now live in
the package. `workflows.ts` is new and is one line.

The split of responsibility is the point: the package owns *how* a graph executes
(traversal, event ordering, the activity contract), the app owns *what* nodes do and
*where* data lands.

## Three entry points, because Temporal has three environments

| Import | Runs in | Why separate |
| --- | --- | --- |
| `@workflowbuilder/temporal` | worker process | the plugin and activities |
| `@workflowbuilder/temporal/workflow` | V8 sandbox | sandbox code cannot touch Node built-ins or the worker package |
| `@workflowbuilder/temporal/client` | backend process | a backend that only starts runs should not install the worker's native binary |

## Decisions worth knowing before reviewing

**`src/core-contract.ts` imports sibling packages by relative path.** `execution-core`
and `types` are private and source-only. Imported by package name, that name survives
into the emitted `.d.ts`, and since neither is on npm consumers would get working JS
and broken types. Relative paths make them project files, so declarations get inlined.
`dts.resolve`, tsconfig `paths` and api-extractor were all tried first. An ESLint rule
blocks `@workflow-builder/*` everywhere else in `src/`.

**`private: true` on a package that has a release workflow.** Without it the package
would ship on its own: its changeset is in `.changeset/`, so the next SDK release bumps
it, `changeset tag` tags it, and one `git push --tags` publishes it. Removal procedure
is at the top of `packages/sdk/RELEASE.md`.

**A behaviour change in `execution-core` now needs a changeset for this package**,
because it is bundled into `dist` and this release is how such a change reaches
consumers. `pr-check` warns (does not fail) when those paths change without one.

**The task queue name lives in one place now.** It used to be written out in both apps.
If those strings drifted, the backend would accept a run, return 202, and the run would
sit in a queue nobody listens to: no error, just silence.

**The cancel fix is unrelated to the refactor** and predates this branch by about a
month. It surfaced during end-to-end testing. Cancelling closed the event stream before
sending the DELETE, so the `execution_cancelled` event the worker emits a moment later
never reached the client and the panel stayed in `running` forever. The stream adapter
already closes the connection on a terminal event, so removing the early disconnect was
the whole fix.

## Verification

249 tests pass (execution-core 124, temporal 25, backend 96, worker 4). The worker's
count drops from 14 because the emitter tests moved into the package.

Two tests are worth knowing about. `bundling.test.ts` runs Temporal's real workflow
bundler against the re-export pattern, which catches a Node built-in sneaking into
`src/workflow/` before a consumer's runtime does. `api-surface.test.ts` pins the exports
of all three entry points plus the default activity timeouts (10m/2 attempts for nodes,
30s/5 for database writes), so an upgrade cannot quietly change how long a node may run.

Beyond CI: a full run through `pnpm dev:ai-studio` including a cancellation, and a check
that the worker's `workflows.ts` bundles through the built `dist` rather than the package
sources, which is the path a real consumer takes. CI also runs `publint` and
`arethetypeswrong`, because three subpaths with separate type files is where an `exports`
map breaks quietly.

## Follow-ups, not blockers

- Replay harness: the package must replay Event Histories recorded by older versions, and
  the histories can only be recorded now that the worker actually runs on it.
  `test/replay/README.md` covers how.
- WB-524 (per-node-type timeouts) lands inside the package, before any release tag.

## Note for reviewers

After editing `packages/temporal/src`, run `pnpm build:temporal` before `pnpm dev:worker`.
The apps import the built `dist`, exactly like `packages/ui`. Noted in `CLAUDE.md`.


dawidaksamski (approved):

lukasz-jazwa:
Two findings from a review against Temporal's partner integration guide — the first one is worth fixing before this code is ever shown to Temporal, the second is a question to settle.

**1. Plugin name doesn't follow Temporal's required format.**
`PLUGIN_NAME` in `packages/temporal/src/plugin.ts` is `@workflowbuilder/temporal`. Temporal's Technical Review Standards require the plugin name (not the npm package name) to follow the format `my_library.MyPlugin` — in our case `workflowbuilder.WorkflowBuilderPlugin`. This name is what shows up in users' logs, and it's also how Temporal attributes Cloud usage telemetry to a partner, so getting it right matters beyond style. One-line fix.

**2. `@temporalio/plugin` is a regular dependency while the other three Temporal packages are peer dependencies.**
In `packages/temporal/package.json`, a consumer pins their own version of `@temporalio/worker` / `workflow` / `client` (peers at `^1.23.0`), but `@temporalio/plugin` gets installed independently and resolves to the newest 1.x. Those versions can drift apart over time, and Temporal generally expects all of its SDK packages to be on one version. Suggestion: make `@temporalio/plugin` a peer dependency like the others, or at least state in the README that all `@temporalio/*` packages must share one version.

Everything else from the review checklist looks good — entry-point split, sandbox lint guards, the bundling test, and the pinned API surface are exactly what the submission needs.

lukasz-jazwa:
Follow-up on point 1 after double-checking the sources — one nuance worth adding.

The written standard and the docs agree on the dotted format: the partner guide's Technical Review Standards say the plugin name "should be of the standard format `my_library.MyPlugin`", and every example in the plugins guide on docs.temporal.io passes `"organization.PluginName"` to `SimplePlugin`.

However, the two reference plugins the guide itself points at don't follow it: Vercel's `@temporalio/ai-sdk` registers `name: 'AiSDKPlugin'` and Pydantic AI registers `name='PydanticAIPlugin'` — bare class names, no prefix. The SDK itself doesn't enforce or transform the name either.

So the practical takeaway stays the same but with more room: the current `'@workflowbuilder/temporal'` (an npm package name) matches none of the observed conventions and should change either way. `workflowbuilder.WorkflowBuilderPlugin` is the safest pick since it satisfies both the written standard and the docs examples; `WorkflowBuilderPlugin` alone would likely also pass review, given that's what the reference implementations do. I'd also retract my earlier claim that the dotted format matters for Cloud usage telemetry — the guide only says plugins give Temporal usage visibility, without tying that to the name format.

lukasz-jazwa (approved):

.github/workflows/pr-check.yml
```diff
@@ -103,6 +104,65 @@ jobs:
...
+
+      - name: Verify the published shape
+        # Three subpaths with their own types — exactly where an exports map breaks
+        # quietly, and consumers would be the ones to find out. Versions pinned so a
+        # tool release cannot redden PRs on its own; the esm-only attw profile matches
+        # the package (ESM without CJS is a finding under stricter profiles).
+        working-directory: packages/temporal
+        run: |
+          pnpm dlx publint@0.3.24
+          pnpm dlx @arethetypeswrong/cli@0.18.5 --pack --profile esm-only
```
librowski, line 151: This could potentially be a good improvement for other WB packages (like `sdk`)

.github/workflows/pr-check.yml
```diff
@@ -103,6 +104,65 @@ jobs:
...
+
+      - name: Changeset guard for bundled execution packages
+        # execution-core and types are private and source-only: they are bundled into
+        # @workflowbuilder/temporal's dist, so a behaviour change in either ships to
+        # consumers only when this package is released. A warning, not a failure —
+        # a pure refactor legitimately needs no changeset.
+        run: |
+          CHANGED=$(git diff --name-only "origin/${{ github.base_ref }}...HEAD")
```
librowski, line 159: I’m not sure a warning-only guard will be effective here. ::warning creates an annotation in the check details, but the job remains green and the warning is easy to miss.

If this is meant to enforce a real release contract, could we make it a small required check that fails when execution-core or types changes without a Temporal changeset? Pure refactors could use an explicit `no-temporal-changeset` label as an escape hatch. The check could run on `opened`, `synchronize`, `reopened`, `labeled`, and `unlabeled`, so changing the label re-evaluates it without rerunning the full build workflow.

piotrblaszczyk: You are right that the warning is easy to miss, and both technical details in your proposal are correct.

One thing to factor in before we switch to failing: main currently has no branch protection (`required_status_checks.contexts` is empty), and the `no-temporal-changeset` label does not exist yet. As things stand, a failing job does not block a merge either, it is just more visible: a red X instead of a yellow annotation. To make this an actual required check we would also need to create the label and enable required status checks on main, and the second one is a policy decision.

My concern with failing plus a label escape hatch is what it converges to. CI cannot judge the thing this guard is about, which is whether a change in execution-core is consumer-visible. Test-only changes, comments and internal refactors are common, so each of them would need the label, and the label would start being applied by reflex. That is worse than a warning, because a green check then only means somebody clicked something.

Let's discuss this further

piotrblaszczyk:
@lukasz-jazwa heads-up: three commits landed after your approval, and one reverses the
dependency change you asked for.

`@temporalio/plugin` went back to `dependencies`, and `client` and `workflow` moved
there with it. Only `worker` stays a peer. Reason: peers do not resolve for a workspace
package in a `--prod` install. pnpm does not link them into a repo-linked package's
`node_modules`, and locally that was masked by its devDependencies. Building the
deployment image showed it (nothing in CI builds it): both apps failed with
`Cannot find package '@temporalio/client'` / `'@temporalio/plugin'`. Worth noting
`client` was already a peer before your comment, so this predates the plugin question.

Dependencies is also what Temporal does with its own plugins. `@temporalio/ai-sdk` and
`@temporalio/interceptors-opentelemetry` both keep the SDK there, neither uses optional
peers.

Your point holds, just enforced differently: we now pin three SDK packages in one place
instead of relying on the consumer to install a matching set. Trade-off is that they no
longer pick those versions, which the README now states.

The other two commits: the Dockerfile builds the package in a separate stage (the
runtime install is `--prod`, so `tsup` was missing for `prepare`), and the release
workflows moved to Node 24 for the OIDC publish. Same fix release-sdk.yml got after a
failed publish.

PR #110, opened 2026-08-31
feat(execution-core): Record step inputs with sensitive data redaction

dawidaksamski:
## Record step inputs on `node_started`, with write-time secret redaction

The `node_started` event carried no payload, so there was no record of what a step
was actually given — blocking failure diagnosis without reproduction, and the future
correct-and-retry feature. Per the ticket, recording inputs required secret redaction
at write time to land in the same change, since event history is immutable.

### What changed

node_started now records { config, visibleNodeIds } — the node's frozen config plus the ids of outputs visible at start. Output values are not copied (each is already recorded once in its own node_completed/node_failed event, avoiding quadratic history growth); the exported reconstructNodeInputs(events, nodeId) joins ids back to values. The worker's emitEvent activity warns when a serialized payload exceeds 512 KB
No DB migration, no backend/SSE changes, no changeset (private packages only).

### Out of scope

Real secrets still cross into Temporal history via `runWorkflow`/`executeNode` activity
args and sit in `workflow_snapshot_json` — pre-existing, unchanged here; fixed by the
Temporal Payload Codec. Value-borne secrets (resolved
`{{variables.x}}` templates) are not caught by key-based redaction — belongs to the
variables feature.


piotrblaszczyk:
One design question before merge, about the size of the recorded snapshot.

`node_started` copies the full wave snapshot into every event. Since executors may read any completed node's output, node k records k-1 outputs, so a run stores ~N²/2 output copies across its `node_started` events. Every one of those outputs is already recorded exactly once in its own `node_completed`, so the copies add no information. The cost lands on four surfaces at once: Postgres event rows, Temporal history (50 MB/run cap, via `emitEvent` activity args), the SSE stream, and the browser store. A 20-node run with 100 KB LLM outputs carries ~19 MB of duplicates next to ~2 MB of real data.

Proposal: record `{ config, visibleNodeIds }` instead of `{ config, nodeOutputs }`, where `visibleNodeIds = Object.keys(context.nodeOutputs)`. The exact inputs stay fully reconstructible: sequence numbers are assigned synchronously and waves are barriered, so "visible at start" is precisely "completion events with a lower sequence". Recording the id list makes that rule explicit rather than implied by the scheduler. A shared `reconstructNodeInputs(events, nodeId)` helper covers the readers, which already load the full event history anyway. Fidelity is identical, since `node_completed.output` passes through the same `redactSensitive`.

Separately, a cheap guard worth adding in the `emitEvent` activity (outside the sandbox): log a warning when a serialized payload exceeds some threshold, so we see runs approaching the Temporal limits before they fail.

piotrblaszczyk:
One more request, about the `WB-597` / `WB-435` references in code comments and the README.

This repo is open source, but the ticket numbers point to our private ClickUp, so for
any external reader they are dead links and a signal that the real roadmap lives
somewhere they can't see. The comments themselves already carry the substance, so the
IDs add nothing for them.

Could you replace the ticket IDs with descriptive follow-up slugs? The convention:

- The comment stays self-sufficient: it states the limitation and the direction of the
  fix in plain words.
- Instead of a ticket number it ends with a stable kebab-case slug naming the work,
  e.g. `(follow-up: temporal-payload-codec)`.
- The ClickUp task gets a matching `Code marker: temporal-payload-codec` line in its
  description, so when we pick the task up later we grep the repo for the slug and land
  on every relevant site. Grep survives file moves; the ID stays in commit messages and
  the PR description, so `git blame` still leads to full context.

Example, in `execution-context.ts`:

    // before
    // ...encrypting that history is WB-597 subtask 4.

    // after
    // ...encrypting that history is planned (follow-up: temporal-payload-codec).

Concretely in this PR: the two WB-597 mentions in `redact.ts`, the one in
`execution-context.ts`, the test comment referencing subtask 5 (suggest
`follow-up: provider-error-details`), and the WB-597 mention in the README section.
The value-based redaction note could use `follow-up: value-based-redaction`.

When you rename them, please add the matching `Code marker:` lines to the ClickUp
tasks so the reverse lookup works from day one.

Also, please codify the rule in `CLAUDE.md` in this same PR, so the next change follows
it automatically instead of relying on review to catch it. Something along these lines:

> **No ticket IDs in code.** Code comments, READMEs, and other in-repo docs must not
> reference ClickUp IDs (`WB-xxx`); the repo is open source and those are dead links
> for external readers. A comment must be self-sufficient. For planned work, end it
> with a stable kebab-case slug: `(follow-up: <slug>)`, and add a matching
> `Code marker: <slug>` line to the ClickUp task. To find the code sites for a task,
> grep the repo for its slug. Ticket IDs stay in commit messages and PR descriptions.

apps/execution-worker/src/engines/temporal/worker.ts
```diff
@@ -30,13 +30,24 @@ const nodeExecutors: NodeExecutorRegistry<AiStudioNode> = {
   'ai-studio/visualize': executeVisualize,
 };
 
+// Mirrors Temporal's own per-blob warn threshold. Payloads this large also count
+// toward the 50 MB per-run history cap (emitEvent args are recorded in history),
+// so a warning here surfaces runs drifting toward the limit before they hit it.
+const PAYLOAD_WARN_BYTES = 512 * 1024;
```
piotrblaszczyk, line 38: please verify it and add link to the temporal docs to have reference in the future

packages/execution-core/src/graph-runner.test.ts
```diff
@@ -1403,14 +1404,15 @@ describe('runGraph — incomplete runs (dead ends)', () => {
...
     const runner = makeRunner({ A: { output: 'a-result' } });
     const events = makeEvents();
 
     await runGraph(makeInput([start('A'), trigger('B')], [edge('e1', 'A', 'B')]), runner.port, events.port);
 
-    expect(startedPayload(events.events, 'A')).toEqual({ config: {}, nodeOutputs: {} });
-    expect(startedPayload(events.events, 'B')).toEqual({ config: {}, nodeOutputs: { A: 'a-result' } });
+    expect(startedPayload(events.events, 'A')).toEqual({ config: {}, visibleNodeIds: [] });
+    expect(startedPayload(events.events, 'B')).toEqual({ config: {}, visibleNodeIds: ['A'] });
+    expect(reconstructNodeInputs(events.events, 'B')).toEqual({ config: {}, nodeOutputs: { A: 'a-result' } });
```
piotrblaszczyk, line 1415: there is still the old data structure

piotrblaszczyk (approved):

piotrblaszczyk (approved):

packages/execution-core/src/graph-runner.test.ts
dawidaksamski: It's the returned type of the `reconstructNodeInputs` 😊

apps/execution-worker/src/engines/temporal/worker.ts
dawidaksamski: Done! 😊

PR #112, opened 2026-09-01
Test/wb 527 replay history harness

piotrblaszczyk:
## What

The repo's first real Temporal test harness: `packages/temporal/test/replay/replay.test.ts` starts an in-memory dev server (`@temporalio/testing`), runs the fan-out graph (`start → (left, right) → join`) through the plugin, and then:

1. asserts the `ActivityTaskScheduled` count per activity type,
2. replays the freshly recorded history through `Worker.runReplayHistory`,
3. replays a committed history (`histories/v0-parallel-wave.json`) recorded before the current code — the cross-version guard.

Regenerate histories with `UPDATE_REPLAY_HISTORIES=1 pnpm --filter @workflowbuilder/temporal test`; `test/replay/README.md` has the rules (including what a red replay means before vs after the first release).

## Why

The existing determinism test in execution-core proves a run is reproducible against itself (re-execution equivalence). It structurally cannot see rule 9 of the replay audit: a change that is self-consistent but breaks runs recorded by older code. Only replaying a real committed history exposes that, as a `DeterminismViolationError` from Temporal's own replayer. Temporal's review standards also ask for exactly this test shape, so we show up with both forms. `replay-audit.md` updated accordingly.

Also fixes pre-commit prettier to respect the root `.prettierignore` from every workspace — the committed histories must stay byte-identical to what `temporal workflow show --output json` emits.

## Verified

Whole suite runs in ~2s on the normal `pnpm test` path. Mutation-checked: one extra `emitEvent` in `runGraph` turns the activity-count assertion and the cross-version replay red, while the same-code replay stays green (which is why the committed history is the guard that matters).

No changeset: tests, docs, and dev tooling only — nothing ships in `dist`. The CI changeset warning for `packages/execution-core/**` is expected (only `replay-audit.md` changed).


dawidaksamski (approved):

lint-staged.config.mjs
```diff
@@ -1,7 +1,17 @@
...
+// path off this module's own location pins it to the root file from every workspace.
+const prettierIgnore = fileURLToPath(new URL('.prettierignore', import.meta.url));
+
 /**
  * @type {import('lint-staged').Configuration}
  */
 export default {
-  '*.{ts,tsx,js,json,css,astro,md,mdx}': (files) => `prettier --write --log-level=silent ${files.join(' ')}`,
+  '*.{ts,tsx,js,json,css,astro,md,mdx}': (files) =>
+    `prettier --write --ignore-path ${prettierIgnore} --log-level=silent ${files.join(' ')}`,
```
librowski: To make it more reliable against paths containing spaces:

```suggestion
    `prettier --write --ignore-path "${prettierIgnore}" --log-level=silent ${files.join(' ')}`,
```

librowski (approved):

PR #116, opened 2026-09-02
Feat/wb 524 per node activity options

piotrblaszczyk:
## What

Node activities were proxied once at module scope with one fixed profile, so nothing about a node could reach its scheduling options. The proxy now happens per call inside the activity runner port (`packages/temporal/src/workflow/run-workflow.ts`), which buys two things:

1. WB-523: each activity is scheduled with the node's authored label as its Temporal Summary, so Event History reads like the diagram instead of a column of identical `executeNode` rows.
2. WB-524: per-node-type `startToCloseTimeout` and `retry.maximumAttempts`, declared as one constant and handed to both `WorkflowBuilderPlugin`, which checks it at `Worker.create`, and `createRunWorkflow`, which checks it, freezes a snapshot and resolves against that.

Every profile check sits in `packages/temporal/src/workflow/profile-validation.ts`; the README section "What the profile check covers" states what it deliberately does not check.

Three changes reach outside the package. `BaseNode` gains an optional `label`, which the reference backend's `mapNode` lifts out of `data.properties` alongside `errorPolicy` and `role`. The AI Studio node types intersect `BaseNode`, so those runner-level fields are declared rather than only present at runtime. And `resolveExecutor` in execution-core looks up through `Object.hasOwn`: a node typed `constructor` used to resolve `Object.prototype.constructor` and have it called as its executor. That last one predates this branch, fixed here because the branch hardened two analogous lookups.

## Why

Both tasks say to do them together and the reason holds: the rebuild is the entire substance, each feature on top of it is a few lines. Split in two, the second PR is "add one field to an options object" and has to branch off the first rather than `main`.

Validation is where the arguing happened. A profile that only fails once a node is scheduled throws inside `executeNode`, the one place the graph's `errorPolicy` can absorb it: with `continue` the run closes as completed, with a configuration error recorded as a business one. So the checks moved up, to `Worker.create` and to workflow-module evaluation, and the snapshot is deep-frozen so a map edited after `createRunWorkflow` cannot slip past. What stays unchecked is bounded on purpose (WB-630, code marker `temporal-profile-wire-validation`): profile values come from a typed constant in consumer source, not from input, and every extra check is another promise that needs a test.

## Verified

Live against a running Temporal, not only unit tests. Four-node graph, three labelled and one deliberately not:

    activityId  3 -> "Receive order"
    activityId  6 -> "Check amount"
    activityId  9 -> "Render summary"
    activityId 12 -> (none, falls back to the activity type)
    run closed: completed

Measured rather than assumed, and each measurement is why a bound sits where it does: 380 bytes of Summary content serializes to a 409-byte payload against the server's 400-byte cap, `maximumAttempts: 4294967296` arrives on the wire as `0` which Temporal reads as unlimited retries, and `'0.0000001ms'` rounds to a zero Duration, the same silent wedge as `'0s'`.

65 tests in `@workflowbuilder/temporal` (26 before), 149 in execution-core, 99 in backend. `apps/docs` typecheck fails identically on `main`, unrelated.

No changeset: `@workflowbuilder/temporal` is still `private: true` and not on npm, so an entry would describe a change against a version nobody installed. `pr-check.yml` skips the guard while that holds.

## Event History reads like the diagram

Screenshot 2026-09-03 at 12 11 15

Customer Support Triage, one run. The decision routed this ticket to the billing
branch, so the two other reply nodes have no rows.


apps/backend/src/domain/mapper/from-integration-data.ts
```diff
@@ -26,24 +26,28 @@ export function mapToExecutionModel(workflowId: string, data: WorkflowSnapshot):
...
-  const { errorPolicy: rawErrorPolicy, ...config } = node.data.properties ?? {};
+  const { errorPolicy: rawErrorPolicy, label: rawLabel, ...config } = node.data.properties ?? {};
   const errorPolicy = isErrorPolicy(rawErrorPolicy) ? rawErrorPolicy : undefined;
+  const label = isNonEmptyString(rawLabel) ? rawLabel.trim() : undefined;
   const role: NodeRole | undefined = node.data.isStartNode === true ? 'start' : undefined;
   return {
     id: node.id,
     type: node.data.type,
     config,
+    ...(label === undefined ? {} : { label }),
```
librowski: We may consider doing it this way

```ts
import { isDefined, pickBy } from 'remeda';

return {
  id: node.id,
  type: node.data.type,
  config,
  ...pickBy({ label, errorPolicy, role }, isDefined),
};
```

apps/execution-worker/src/domain/ai-studio-nodes.ts
```diff
@@ -2,6 +2,11 @@
 // not by execution-core or @workflow-builder/types — those layers only know
 // the generic BaseNode shape. A different product would define its own union
 // here and register matching executors.
+import type { BaseNode } from '@workflowbuilder/temporal';
+
+// Intersected with BaseNode so the runner-level fields the backend lifts out of
+// `data.properties` (`label`, `errorPolicy`, `role`) are declared, not just present.
```
librowski: I'd remove mentioning implementation details so the comment doesn't outdate as soon as it happens

librowski (approved):

PR #121, opened 2026-09-07
[DS 2.0] Use the measured source-node height for self-loop edges

librowski:
## Problem

`LabelEdge` computed a source-node height to place the label of a self-connecting edge but did not pass it to `SelfConnectingEdge`, whose `nodeHeight` prop defaulted to `0`: the loop was drawn as if the node had no height while the label was positioned as if it had. The height itself was read from `sourceNode.height`, which xyflow 12 sets only when a change explicitly requests attributes - normal ResizeObserver updates write `measured.height`.

A first version read the height through `getNode()`; a browser check (review of 8.09) showed that a snapshot read does not follow a later remeasurement: after adding a branch to a Decision node the loop and label stayed at the old height until an unrelated rerender.

## Change

- The source node's height is read from the React Flow store through a shared `useSelfLoopNodeHeight(source, target)` hook (exported): `measured.height ?? height ?? 0`, `0` for non-self edges. A one-line comment records the xyflow fact behind the order (`measured.height` on every remeasure, `height` only when attributes are set explicitly).
- `LabelEdge` passes the value to `SelfConnectingEdge` so loop and label stay in lockstep. `SelfConnectingEdge` itself now reads the same hook when `nodeHeight` is omitted, so a custom edge that delegates to it (as `dashed-edge.tsx` suggests) no longer draws the loop for a zero-height node; the prop's JSDoc says when to pass a value.
- Specs: measured height, fallback to an explicit `height` on a realistic `{ measured: {}, height }` node, unknown node, a regular edge never touching `nodeLookup`, and `SelfConnectingEdge` with and without `nodeHeight`.
- Changeset: `@workflowbuilder/sdk` minor (`SelfConnectingEdge` changes its default behaviour; `useSelfLoopNodeHeight` is a new export).

Not covered by an automated test: the live remeasurement itself (the spec exercises the selector with a mocked store; a real React Flow harness would be a new test layer).

## Out of scope, to settle with design

`SELF_CONNECTING_EDGE_LABEL_OFFSET` is documented as the distance from the node's top edge, but the loop is offset from the port by `nodeHeight + 100`, so the apex sits `nodeHeight / 2 + 100` above the top when ports are vertically centred. Whether that geometry is the requirement is a design question; the constant and its doc are unchanged here.

Edge corner radius has three values today: 16px in code (`EDGE_CURVE_RADIUS`), 2px in the token export (`canvas-edge-corner-radius`), 18px in the Figma vectors. Adopting the export would visibly straighten every corner; not changed here.

### DIFF
--- a/packages/sdk/src/features/diagram/edges/label-edge/label-edge.spec.tsx
+++ b/packages/sdk/src/features/diagram/edges/label-edge/label-edge.spec.tsx
@@ -1,0 +1,78 @@
+import { render, screen } from '@testing-library/react';
+import type { EdgeProps } from '@xyflow/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type { WorkflowBuilderEdge } from '../../../../node/node-data';
+import { LabelEdge } from './label-edge';
+
+const { nodeLookup } = vi.hoisted(() => ({ nodeLookup: new Map<string, unknown>() }));
+
+vi.mock('@xyflow/react', () => ({
+  getSmoothStepPath: () => ['M 0 0', 0, 0],
+  useStore: (selector: (state: { nodeLookup: Map<string, unknown> }) => unknown) => selector({ nodeLookup }),
+}));
+
+vi.mock('../edge-label-renderer/edge-label-renderer', () => ({
+  EdgeLabel: ({ id, labelY }: { id: string; labelY: number }) => (
+    <span data-testid="edge-label" data-edge-label-id={id} data-label-y={labelY} />
+  ),
+}));
+
+vi.mock('../enhanced-base-edge/enhanced-base-edge', () => ({
+  EnhancedBaseEdge: ({ id, path }: { id: string; path: string }) => (
+    <svg>
+      <path data-edge-id={id} d={path} />
+    </svg>
+  ),
+}));
+
+vi.mock('./use-label-edge-hover', () => ({
+  useLabelEdgeHover: () => ({
+    style: {},
+    hovered: false,
+    onMouseEnter: vi.fn(),
+    onMouseLeave: vi.fn(),
+  }),
+}));
+
+const selfConnectingEdgeProps = {
+  id: 'self-loop',
+  source: 'node-1',
+  target: 'node-1',
+  sourceX: 100,
+  sourceY: 300,
+  targetX: 200,
+  targetY: 300,
+  sourcePosition: 'right',
+  targetPosition: 'left',
+  data: { label: 'Loop' },
+} as EdgeProps<WorkflowBuilderEdge>;
+
+beforeEach(() => {
+  nodeLookup.clear();
+});
+
+describe('LabelEdge', () => {
+  it('uses the measured source height for self-loop geometry and label placement', () => {
+    nodeLookup.set('node-1', { measured: { height: 80 } });
+
+    const { container } = render(<LabelEdge {...selfConnectingEdgeProps} />);
+
+    expect(container.querySelector('[data-edge-id="self-loop"]')?.getAttribute('d')).toContain('Q 125 120 109 120');
+    expect(screen.getByTestId('edge-label').dataset.labelY).toBe('120');
+  });
+
+  it('falls back to the explicit node height when no measurement exists', () => {
+    nodeLookup.set('node-1', { height: 40 });
+
+    render(<LabelEdge {...selfConnectingEdgeProps} />);
+
+    expect(screen.getByTestId('edge-label').dataset.labelY).toBe('160');
+  });
+
+  it('treats an unknown source node as zero height', () => {
+    render(<LabelEdge {...selfConnectingEdgeProps} />);
+
+    expect(screen.getByTestId('edge-label').dataset.labelY).toBe('200');
+  });
+});
--- a/packages/sdk/src/features/diagram/edges/label-edge/label-edge.tsx
+++ b/packages/sdk/src/features/diagram/edges/label-edge/label-edge.tsx
@@ -1,4 +1,4 @@
-import { type EdgeProps, getSmoothStepPath, useReactFlow } from '@xyflow/react';
+import { type EdgeProps, getSmoothStepPath, useStore as useReactFlowStore } from '@xyflow/react';
 
 import { Icon } from '@workflow-builder/icons';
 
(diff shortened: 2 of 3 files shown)


packages/sdk/src/features/diagram/edges/label-edge/label-edge.tsx
```diff
@@ -34,7 +34,11 @@ export function LabelEdge({
   source,
   target,
 }: EdgeProps<WorkflowBuilderEdge>) {
-  const { getNode } = useReactFlow();
+  const nodeHeight = useReactFlowStore((state) => {
+    if (source !== target) return 0;
+    const node = state.nodeLookup.get(source);
+    return node?.measured?.height ?? node?.height ?? 0;
```
piotrblaszczyk: Could you add a one-liner here recording the xyflow fact? Something like:

`// xyflow writes measured.height on every remeasure; height only when a resize sets attributes.`

The PR description explains it well, but that text does not ship, and the next PRs in the stack touch this exact area. This is the "external behaviour with no visible trace" case from CLAUDE.md § Code comments.

packages/sdk/src/features/diagram/edges/label-edge/label-edge.tsx
```diff
@@ -83,6 +85,7 @@ export function LabelEdge({
           target={target}
           sourcePosition={sourcePosition}
           targetPosition={targetPosition}
+          nodeHeight={nodeHeight}
```
piotrblaszczyk, line 84: Passing it here fixes this caller, but `nodeHeight` still defaults to `0` on the public `SelfConnectingEdge` (`self-connecting-edge.tsx:9` and `:62`), so any other caller silently gets the pre-PR geometry. There is no other caller in-repo today, but the demo's `dashed-edge.tsx` (lines 19-20) explicitly points custom-edge authors at that component without mentioning the prop.

Cheapest fix for this PR: a JSDoc line on the prop saying what to pass (`measured.height` from the React Flow store) and that omitting it draws the loop as if the node had no height.

Nicer follow-up: derive it inside `SelfConnectingEdge` from `source` when the prop is omitted, using the same selector. `LabelEdge` would keep passing its own value so loop and label stay in lockstep.

packages/sdk/src/features/diagram/edges/label-edge/label-edge.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+  sourcePosition: 'right',
+  targetPosition: 'left',
+  data: { label: 'Loop' },
+} as EdgeProps<WorkflowBuilderEdge>;
+
+beforeEach(() => {
+  nodeLookup.clear();
+});
+
+describe('LabelEdge', () => {
```
piotrblaszczyk, line 55: Small gap, non-blocking: nothing exercises the `source !== target` branch. A test that renders a normal edge (`source: 'a', target: 'b'`) and asserts the selector never touches `nodeLookup` (spy on the hoisted map's `get`) would guard the early exit for the PRs stacked on this one.

packages/sdk/src/features/diagram/edges/label-edge/label-edge.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    nodeLookup.set('node-1', { measured: { height: 80 } });
+
+    const { container } = render(<LabelEdge {...selfConnectingEdgeProps} />);
+
+    expect(container.querySelector('[data-edge-id="self-loop"]')?.getAttribute('d')).toContain('Q 125 120 109 120');
+    expect(screen.getByTestId('edge-label').dataset.labelY).toBe('120');
+  });
+
+  it('falls back to the explicit node height when no measurement exists', () => {
+    nodeLookup.set('node-1', { height: 40 });
```
piotrblaszczyk: Nit: a real `InternalNode` always has `measured: {}` (set by `adoptUserNodes`), with `height` undefined before measurement. `{ measured: {}, height: 40 }` would mirror the real shape and show that the fallback is reached with `measured` present but unpopulated.

packages/sdk/src/features/diagram/edges/label-edge/label-edge.tsx
```diff
@@ -65,8 +69,6 @@ export function LabelEdge({
   };
 
   if (source === target) {
-    const sourceNode = getNode(source);
-    const nodeHeight = sourceNode?.height ?? 0;
     const selfConnectingLabelY = sourceY - (nodeHeight + SELF_CONNECTING_EDGE_LABEL_OFFSET);
```
piotrblaszczyk, line 68: FYI, pre-existing, no action in this PR: the JSDoc on `SELF_CONNECTING_EDGE_LABEL_OFFSET` (`edge.consts.ts:16-19`) says the offset is measured from the node's top edge, but `createSelfConnectingPath` offsets from the port Y by `nodeHeight + 100`, and so does this line. If handles sit at the vertical centre of the node, the apex is `nodeHeight / 2 + 100` above the top, not `100`. Worth knowing before the node-geometry PR (#123) takes that text at face value.

piotrblaszczyk (approved):

packages/sdk/src/features/diagram/edges/label-edge/label-edge.tsx
librowski: Added, in the shared `useSelfLoopNodeHeight` hook the selector moved into: `// xyflow writes measured.height on every remeasure; height only when a resize sets attributes.`

packages/sdk/src/features/diagram/edges/label-edge/label-edge.tsx
librowski: Went with the nicer version: `SelfConnectingEdge` now reads the measured height through `useSelfLoopNodeHeight` when `nodeHeight` is omitted; `LabelEdge` keeps passing its own value so loop and label stay in lockstep. The hook is exported for custom edges, the prop's JSDoc says when to pass a value, and the changeset moved to minor since the default behaviour of a public component changes. Two specs cover both paths.

packages/sdk/src/features/diagram/edges/label-edge/label-edge.spec.tsx
librowski: Added: a regular edge asserts `nodeLookup.get` is never called and the label Y falls back to the smooth-step midpoint.

packages/sdk/src/features/diagram/edges/label-edge/label-edge.spec.tsx
librowski: Changed to `{ measured: {}, height: 40 }`.

packages/sdk/src/features/diagram/edges/label-edge/label-edge.tsx
librowski: Confirmed: the apex sits `nodeHeight / 2 + 100` above the top edge when ports are centred (measured 132px on a 64px node). Left the constant and its doc untouched here and recorded it as a design question (is that geometry the requirement, or the documented 100 from the top edge); the PR description now lists it under out of scope.

PR #122, opened 2026-09-07
[DS 2.0] Render the default canvas port at the designed 8px

librowski:
## Problem

The default port declares content 4px + border 2px per side with `box-sizing: content-box`, intending an 8px outer size. The base stylesheet of `@xyflow/react` 12.10.0 imposes `min-width: 5px; min-height: 5px` on handles; with content-box the minimum applies to the content, so the effective size was 9x9px. Hover (content 12px) was already correct at 16px.

## Change

- `handle.module.css`: `min-width: 0; min-height: 0` on the base handle rule.
- `built-css-pitfalls.md` documents the inherited-minimum trap and how to check it by hand. A dist check for the handle geometry was added and then removed in review: it matched every rule ending in ` .react-flow__handle`, so any later state rule on the bare handle would have failed it, and exact-string matching against minified output was fragile.
- Changeset: `@workflowbuilder/ui` patch.

## Verification

Build, `check:built-css` (prefixes, root-scoped public variables, imports, asset URLs), ui tests and stylelint pass. Manual browser pass in the demo (Simple Workflow, 30 handles, layout px via `offsetWidth`): default and `connectionindicator` 8x8, handle hover 16x16, node hover indicators 16x16, `connectingfrom` 8x8 with crosshair, top/bottom handle rules only reposition (size unchanged). jsdom cannot measure layout, so the automated part is the built-CSS contract.

## Follow-up in this PR: target port while connecting

xyflow assigns `connectingto` to the closest handle within `connectionRadius`, so the pointer is often not over the target and `:hover` cannot express the state; the target needs its own rule. `.connectingto.valid` shares the `.connectingfrom` rule: a target accepted by `isValidConnection` grows to the 16px active port, a rejected one keeps 8px while the connection line reports invalid, other ports stay 8px, drop still creates the edge (verified in the demo with the built package). Design has Default and Hover port states only; design is asked to confirm Hover for the valid target and to say whether an invalid target needs a state of its own. Changeset: `@workflowbuilder/ui` patch.


packages/ui/src/components/node/node-panel/handle.module.css
```diff
@@ -71,8 +73,11 @@
 
     /* State rules repeat the base rule's content-box opt-out: the build
        injects border-box per rule, so an undeclared box-sizing here would
-       shrink the handle in exactly these states. */
-    :global(.connectingfrom) {
+       shrink the handle in exactly these states. The target port needs its
+       own rule: React Flow captures the pointer on the dragged handle, so
```
piotrblaszczyk: Two things about this comment. It now runs five lines, and CLAUDE.md caps comments at three. And the pointer-capture explanation does not match xyflow: `XYHandle.onPointerDown` only registers document-level `mousemove` / `mouseup` listeners, there is no `setPointerCapture` anywhere in `@xyflow/system`.

The stronger reason is simpler: `connectingto` is assigned by proximity. `getClosestHandle` picks any handle within `connectionRadius` (20px by default), so the pointer often is not over the target at all and `:hover` cannot represent it.

Suggest keeping the original three-line box-sizing note and adding one line:

`/* .connectingto is assigned by proximity (connectionRadius), so :hover cannot cover the target. */`

packages/ui/src/components/node/node-panel/handle.module.css
```diff
@@ -71,8 +73,11 @@
...
+       :hover never reaches the node under the cursor while connecting. */
+    :global(.connectingfrom),
+    :global(.connectingto) {
```
piotrblaszczyk: One semantic wrinkle: xyflow sets `connectingto` on whichever handle it snapped to, valid or not. In `isValidHandle`, `toHandle` is filled independently of `isValid`, and `valid` is a separate class on the handle. With this rule a target rejected by `isValidConnection` (a public `WorkflowBuilder.Root` prop) also grows to the active look, which reads as droppable while the connection line reports invalid.

If that is not intended, `:global(.connectingto.valid)` limits the grow to valid targets with a one-token change. If design wants snap feedback regardless of validity, a word here or in the changeset would make it a decision rather than an accident. Either way, could the design question also cover the invalid-target case?

packages/ui/scripts/check-built-css.ts
```diff
@@ -92,6 +100,18 @@ for (const { path: directory, ownsLayerContract } of directories) {
     }
     root.walkAtRules('import', (atRule) => add(file, atRule, 'Built CSS import'));
 
+    root.walkRules((rule) => {
+      if (!rule.selectors.some((selector) => selector.endsWith(' .react-flow__handle'))) return;
```
piotrblaszczyk: `endsWith(' .react-flow__handle')` catches any rule that targets the bare handle, not just the geometry rule. Today only the base rule matches, but a later `.some-state .react-flow__handle { opacity: .5 }` would fail with "Incorrect built handle geometry" even though geometry is untouched, and repeating the six declarations to appease the check would be wrong. The next PRs in this stack add node-state rules, so this will bite soon.

Suggest narrowing to rules that declare `width` (or `box-sizing: content-box`) on that selector, plus asserting that at least one such rule exists so a selector rename cannot silently disable the check.

Smaller note: esbuild already reshapes rules (the two hover rules are merged into one selector list in the SDK dist), so exact-string matching of `border: var(...) solid var(...)` against minified output is a soft spot. Fine for now, worth knowing.

piotrblaszczyk (approved):

packages/ui/src/components/node/node-panel/handle.module.css
librowski: Taken verbatim: the original three-line box-sizing note plus your one-liner about `connectionRadius`. You are right about pointer capture, there is none in `@xyflow/system`; my explanation was wrong.

packages/ui/src/components/node/node-panel/handle.module.css
librowski: Narrowed to `.connectingto.valid`: a target rejected by `isValidConnection` keeps 8px while the connection line reports invalid. Changeset says so. The design question now covers the invalid target as well (Node_port has Default and Hover only).

packages/ui/scripts/check-built-css.ts
librowski: Removed the check instead of narrowing it. It guarded one regression that is already fixed and commented at the source, matched every rule on the bare handle, and compared exact strings against minified output. `built-css-pitfalls.md` keeps the entry with a manual check (`offsetWidth` of a resting handle reads 8) and says why the automated version was dropped.

PR #123, opened 2026-09-07
[DS 2.0] Move the node shell to the DS 2.0 geometry and derive ConnectableItem width

librowski:
Two commits, reviewed together because the second is a consequence of the first.

## 1. Node shell geometry (`@workflowbuilder/ui`, patch)

`node-panel.module.css`: `--wb-public-node-width` 16.125rem (258px) -> `241px`, the design master, expressed in px because it is a canvas dimension (the canvas zoom scales the transform, not the font size). `--wb-public-node-padding` and `--wb-public-node-gap` bind to the node head roles from the token export, `--wb-ds-canvas-node-head-h-pad` and `--wb-ds-canvas-node-head-gap` (8px each, confirmed by design as the master values), instead of the generic `space-100` step. Rendered padding and gap are unchanged; the exported `canvas-node-inset` / `gap-inner` roles (16/12px) were confirmed as orphans and are not used. Public variable names are unchanged; consumer overrides keep working. Start, Workflow and AI templates keep their own `--wb-public-node-gap: 0`.

Cold-load check before the change: the initial viewport fit uses DOM-measured sizes, not the `measured` values stored in fixtures (fit differs between 258 and 241 exactly as the narrower bounds predict, nodes render at 241 from the first frame, transform stable afterwards). The visible two-stage fit on template selection is pre-existing (`template-selector` calls `fitView()` after the initial fit) and width-independent.

## 2. ConnectableItem width (`@workflowbuilder/sdk`, patch)

The old cap `width - 6 * padding + 2 * hpad + 2 * border` used an undocumented factor that only approximated the real horizontal insets (48px subtracted vs 40px real), and its `+ padding/border` terms assumed content-box while the SDK applies a global border-box reset - at 241px the old rule allows 219px where a section offers 201px, so a long label could overflow its section by 18px.

New rule: `width - 2 * (node padding + node border) - 2 * var(--wb-sdk-connectable-item-inset)`, where each wrapping container declares the inset it adds: `NodeSection` (padding + border) for Decision branches and `NodeInfoWrapper` (0.625rem + 1px) for AI tool rows, both giving 201px. The cap narrows (219px -> 201px at 241px) so labels truncate at the section edge instead of overflowing it; the changeset says so. The variable is internal (`--wb-sdk-*`), not a public extension point. Rationale in `connectable-item-width.decision-log.md` (`DECISION-LOGS.md` regenerated); the pending design work is named there with the slug `connectable-item-design-width` per the repo's follow-up convention.

## Verification

Build, `check:built-css`, ui and sdk tests, typecheck, stylelint. Visual pass in Demo (Simple Workflow, Call Flow, User Registration) and AI Studio (Support Triage, AI Debate), light and dark, fitView and 100%: every node 241px, shell padding 8px, no viewport jump after reload, no clipped titles, Decision branch rows exactly 201px with the section inset applied. Not covered: execution markers (need a run).

Known, pre-existing and out of scope: a Decision node in the `DOWN` layout lays its branches out in one horizontal row and `min-width: max-content` lets that row widen the node (621px measured with three branches). Whether the fixed 241px rule covers `DOWN` is a design decision tracked separately.


.changeset/connectable-item-width.md
```diff
@@ -0,0 +1,5 @@
+---
+'@workflowbuilder/sdk': patch
+---
+
+Decision branch rows and AI tool rows derive their width cap from the node shell and their container insets, so long labels keep the full available width after the node shell spacing change instead of truncating early.
```
piotrblaszczyk: The changeset reads as if the cap got wider, but the fix narrows it (219px -> 201px at a 241px shell, 236px -> 218px at the old 258px), so labels stop overflowing the section rather than "keeping the full available width". Consumers reading the CHANGELOG would expect the reverse behaviour.

Suggest: "Decision branch rows and AI tool rows cap their width to the space their section actually offers, so long labels truncate at the section edge instead of overflowing it."

packages/sdk/src/features/diagram/nodes/components/connectable-item/connectable-item-width.decision-log.md
```diff
@@ -0,0 +1,68 @@
...
+
+- The `6 *` factor is not documented anywhere. It only approximates the real horizontal insets
+  between the node edge and the item: shell padding plus border on both sides (2 x 9px) and the
+  section padding plus border on both sides (2 x 11px), 40px in total against the 48px it subtracts.
+- The `+ 2 * padding + 2 * border` terms assume content-box sizing. The SDK applies a global
+  `box-sizing: border-box` reset (`packages/sdk/src/index.css`), so `max-width` already refers to
+  the border box and the terms inflate the cap: at 241px the old rule allows 219px while only 201px
+  are available inside a section, so a long label could overflow its section by 18px.
+
+The design system does not specify a width for these items (register: node geometry gaps). Any cap
```
piotrblaszczyk: "register: node geometry gaps" here and "DS 2.0 divergence register" on line 67 live outside the repo, so for external readers of a public repo these are dead pointers, and this log is indexed from the root `DECISION-LOGS.md`. CLAUDE.md § "Referencing future work" is written for code and READMEs, but the reasoning applies equally here.

Could you state in one sentence what the register records (design has no width for this row, the cap is provisional) and end with a grep-able slug, e.g. `(follow-up: connectable-item-design-width)`? The ClickUp task then carries `Code marker: connectable-item-design-width`. Three more PRs in this stack will likely cite the same register, so settling the wording now saves repeating it.

piotrblaszczyk (approved):

.changeset/connectable-item-width.md
librowski: Replaced with your wording; the cap narrows, labels truncate at the section edge instead of overflowing.

packages/sdk/src/features/diagram/nodes/components/connectable-item/connectable-item-width.decision-log.md
librowski: Both pointers are gone. The log states that design specifies no width for these rows and that the cap is provisional, and ends with `(follow-up: connectable-item-design-width)`; the ClickUp task carries the matching `Code marker` line.

PR #124, opened 2026-09-07
[DS 2.0] Strip measured node sizes from saved and exported diagrams

librowski:
## Problem

`getStoreDataForIntegration` (used by the localStorage, API and props integrations and by export) stripped only `selected` from nodes, so the runtime `measured` sizes travelled with the diagram. Observed in the demo: after the node shell width changed, Save still wrote the previous `measured.width`, so persisted data carried sizes that no longer matched the rendered nodes. Nothing in the SDK reads `measured` back from stored data - nodes are measured again on load.

## Change

- `skipDynamicValuesInNodes` drops `measured` alongside resetting `selected`; JSDoc of `getStoreDataForIntegration` lists measured sizes among the stripped runtime values.
- New `dynamic-values.spec.ts` covers nodes (measured dropped, selection reset, input untouched) and edges (routing points and selection reset).
- Changeset: `@workflowbuilder/sdk` patch (export format no longer contains `measured`).
- Second commit: the 82 stored `measured: { width: 258, height: ... }` entries are removed from the Demo templates and AI Studio flows (9 files). They were dead data after the first commit and still recorded the pre-DS 2.0 width; design confirmed stored sizes are a cache, not a source.


packages/sdk/src/store/slices/diagram-slice/utils/dynamic-values.ts
```diff
@@ -1,7 +1,7 @@
 import type { WorkflowBuilderEdge, WorkflowBuilderNode } from '../../../../node/node-data';
 
 export const skipDynamicValuesInNodes = (nodes: WorkflowBuilderNode[]): WorkflowBuilderNode[] => {
-  return nodes.map((node) => ({
+  return nodes.map(({ measured: _measured, ...node }) => ({
```
piotrblaszczyk, line 4: `dragging` is written by React Flow on every position change (`applyNodeChanges`, `case 'position'`) and is documented as "whether the node is currently being dragged", so it is the same kind of runtime value as `measured`. Nothing in the SDK or the apps reads it, and the 9 fixtures this PR just cleaned still carry 53 `dragging: false` entries.

Since the JSDoc now promises that runtime-only values are stripped, could we drop it in the same destructure (`{ measured: _measured, dragging: _dragging, ...node }`), add one `not.toHaveProperty('dragging')` to the spec, and sweep the `dragging: false` lines out of the fixtures in the second commit? Same shape of change, and it saves a follow-up that would re-touch exactly these files under two stacked PRs.

packages/sdk/src/store/slices/diagram-slice/actions.ts
```diff
@@ -88,9 +88,9 @@ type GetStoreDataParams = {
  * persistable payload to the host — e.g. inside a `props`-strategy
  * `onDataSave` callback or before posting to a custom backend.
  *
- * Dynamic, runtime-only values (selection, computed avoid-edge points,
- * …) are stripped by default; pass `shouldSkipDynamicValues: false` if
- * you specifically need the live values.
+ * Dynamic, runtime-only values (selection, measured node sizes, computed
```
piotrblaszczyk, line 91: The new spec covers the helpers nicely. Could we also add two cases to `actions.spec.ts` for `getStoreDataForIntegration` itself: the default strips `measured` (and `selected`, routing points), and `shouldSkipDynamicValues: false` returns the live objects? It is the function the changeset and this JSDoc name, and today every existing spec mocks it away, so the public contract has no test at the boundary.

piotrblaszczyk (approved):

packages/sdk/src/store/slices/diagram-slice/utils/dynamic-values.ts
librowski: Agreed, same class of value. `getNodesWithoutDynamicValues` now strips `dragging` next to `measured` and `selected`, the bundled Demo and AI Studio templates lost their `dragging: false` lines, and the changeset mentions it. Fixed in #150.

packages/sdk/src/store/slices/diagram-slice/actions.ts
librowski: Added two cases to `actions.spec.ts`: the default output has no `measured`, `dragging` or selection and empty routing points; `shouldSkipDynamicValues: false` returns the live store arrays by reference. Fixed in #150.

PR #126, opened 2026-09-07
[DS 2.0] Keep node title and subtitle on one line with ellipsis and tooltip

librowski:
## Problem

Design rule for DS 2.0 canvas nodes (confirmed by design on the node geometry questions): fixed 241px width, no resize, height grows; title, subtitle and body rows are one line with an ellipsis and the full text in a tooltip; rich body wraps. In code only the title truncated. A long subtitle wrapped, and in the Decision template (`min-width: max-content` on the body) it widened the whole node - observed at 284px in the AI Studio "Customer Support Triage" flow.

## Change

- `node-description.module.css`: the subtitle gets the same one-line ellipsis treatment as the title; the block gets `min-width: 0` and `contain: inline-size`, so its text no longer contributes intrinsic width to a `max-content` body. Nodes keep the design width and grow in height only.
- `NodeDescription` sets the native `title` attribute on both lines so the full text stays reachable (the design system tooltip component can replace it later without changing the contract).
- `ConnectableItem` (Decision branch rows, AI tool rows) gets the same `title` attribute on its label, so every clipped line in the node shell is readable on hover; spec and `@workflowbuilder/sdk` patch changeset.
- Spec covers the tooltip attributes; docs page gains a "Truncation" section; changeset `@workflowbuilder/ui` patch.

## Verification

ui tests (69), sdk ConnectableItem spec, typecheck, build, `check:built-css`, stylelint, eslint. End-to-end pass of the built stack (2026-09-09): 241px on every node in all demo templates and AI Studio, Decision rows 201px, clipped titles and rows carry the full text.


packages/ui/src/components/node/node-description/node-description.module.css
```diff
@@ -11,15 +11,21 @@
     align-items: flex-start;
     overflow: hidden;
     width: 100%;
+    min-width: 0;
+    contain: inline-size;
```
piotrblaszczyk, line 15: `contain: inline-size` zeroes the block's intrinsic contribution, so `NodeDescription` now only renders text when an ancestor has a definite inline size. Inside `NodePanel.Header` it does, via the 241px `.container`, and every render site in the repo is safe. But it is a public ui export: a consumer dropping it into an `inline-flex` or `fit-content` wrapper would see it collapse to 0 with no error. Could we state that constraint in the Truncation section and in the ui changeset?

.changeset/connectable-item-tooltip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@workflowbuilder/sdk': patch
+---
+
+Decision branch rows and AI tool rows show the full label in a tooltip when the text is truncated.
```
piotrblaszczyk, line 5: Nit on wording, since this becomes the CHANGELOG: the `title` is set always, not only when clipped. Suggest: "Decision branch rows and AI tool rows expose the full label as a native tooltip (`title`), so a clipped row stays readable."

packages/ui/src/components/node/node-description/node-description.tsx
```diff
@@ -11,8 +11,12 @@ export type NodeDescriptionProps = {
 export function NodeDescription({ label, description, className }: NodeDescriptionProps) {
   return (
     <div className={clsx(styles['container'], className)}>
-      <span className={clsx('wb-text-title-s-emphasized', styles['title'])}>{label}</span>
-      <span className={clsx('wb-text-node-s', styles['subtitle'])}>{description}</span>
+      <span className={clsx('wb-text-title-s-emphasized', styles['title'])} title={label}>
```
piotrblaszczyk, line 14: The description says the DS `Tooltip` will replace `title` later. Could the tsx carry a one-liner with a follow-up slug, e.g. `// Native title until the DS Tooltip takes over; keyboard users get no tooltip today (follow-up: node-text-ds-tooltip)`? That is the CLAUDE.md convention for future work and keeps it grep-able for the PRs stacked on this one.

piotrblaszczyk (approved):

packages/ui/src/components/node/node-description/node-description.module.css
librowski: Documented rather than changed: the Truncation section in `node-description.mdx` and the changeset now state that the block needs an ancestor with a definite inline size and collapses to zero width inside `inline-flex` or `fit-content` wrappers. The behaviour itself stays, because it is what keeps the node at the design width. Fixed in #150.

packages/ui/src/components/node/node-description/node-description.tsx
librowski: Left as is for now. Whether the DS Tooltip replaces `title` here at all is a design question that is still open (the Figma spec has no Tooltip on these rows), so I do not want to plant a follow-up slug before that lands. Will add the marker together with the decision.

.changeset/connectable-item-tooltip.md
librowski: Taken verbatim. Fixed in https://github.com/synergycodes/workflowbuilder/pull/150.

packages/ui/src/components/node/node-description/node-description.tsx
piotrblaszczyk: Understood that the DS Tooltip is not decided, and the marker does not need to name it. It records a limitation the code has today: keyboard and touch users get no tooltip at all. Something neutral like `// Native title only: keyboard and touch users get no tooltip (follow-up: node-text-tooltip-a11y)` keeps that grep-able whatever design decides. One more thing: the "Not changed" bullet in #150 points at #136 and connectable items; this thread is #126 and `NodeDescription`.

PR #128, opened 2026-09-09
docs(temporal): event history limits, secrets and the tool-loop decision adjustments

jimmeryn:
Three documentation items from the Temporal audit. Docs and comments only, no
runtime change.

- What Event History records, and why secrets belong in the worker's environment
  rather than in `variables`.
- The 2 MB payload and 50 MB history limits, how a chain of N nodes spends them,
  and what an oversize payload actually does.
- The AI agent's tool loop runs inside one activity, so a retry re-runs the whole
  loop. Now owned as a decision instead of read as an accident.

> [!NOTE]  
>**Why READMEs and not the docs site?**
>There is no Temporal page to put these on.
That page is its own audit item, the "Using Workflow Builder with Temporal"
guide. Limits also belong next to the API they constrain. The one docs-site
change is the variable picker guide, which called `variables` a place for secrets.


apps/execution-worker/README.md
```diff
@@ -26,14 +26,15 @@ Requires Postgres + Temporal running. Start them with `pnpm infra:up`.
...
 
-| Var                  | Purpose                            | Default                                              |
-| -------------------- | ---------------------------------- | ---------------------------------------------------- |
-| `OPENROUTER_API_KEY` | AI agent activities (**required**) | —                                                    |
-| `DATABASE_URL`       | Execution events + status          | `postgresql://wb:wb@127.0.0.1:5432/workflow_builder` |
-| `TEMPORAL_ADDRESS`   | Temporal server address            | `127.0.0.1:7233`                                     |
-| `AI_MODEL`           | OpenRouter model ID                | `anthropic/claude-3.5-haiku`                         |
+| Var                  | Purpose                             | Default                                              |
+| -------------------- | ----------------------------------- | ---------------------------------------------------- |
+| `OPENROUTER_API_KEY` | AI agent activities (**required**)  | —                                                    |
```
dawidaksamski: This will be deprecated once https://github.com/synergycodes/workflowbuilder/pull/113 is merged

dawidaksamski (approved):

PR #133, opened 2026-09-10
Feat/wb 501 decision endpoint

piotrblaszczyk:


apps/backend/src/routes/decision.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    const validated = validateSubmittedDecision(found.request, submitted);
+    if (validated.error !== undefined) {
+      return refuse(c, 'decision_invalid', undefined, { details: [validated.error] });
+    }
+    const { decision, action } = validated;
+
+    const waits = await countNodeWaits(executionId, nodeId);
+    if (waits === 0) return refuse(c, 'node_never_parked', nodeId);
+    if (waits !== attempt) return refuse(c, 'attempt_mismatch', undefined, { attempt: waits });
```
dawidaksamski: The attempt check lives here in the backend, but the engine never sees the attempt: `resolveNode` gets only `nodeId` + `resolution`, and the `waits` map in `run-workflow.ts` is keyed by node id alone. So the check isn't atomic with delivery.

Today that's fine (a node parks once, `rerun-source` is 501), but once the rerun loop lands, a count read just before a re-park can deliver into the wrong wait, and the engine has no way to notice. The ticket keys the wait instance on execution + node + attempt for exactly this reason. Decision-log 15 acknowledges the gap, but nothing in code marks the seam.

Two options:

**Preferred** (small, additive, replay-safe since validators write nothing to history):

- add an optional `attempt` to `resolveNode` on the port and to the update input
- give `NodeWaitState` an `attempt` counter, incremented at the park site
- have the validator reject `verdict.attempt !== state.attempt` with a new `verdict_attempt_mismatch` code, mapped to `decision_attempt_mismatch` in `ENGINE_REFUSALS`

The completeness tests will catch any missed dictionary. Old callers keep working since the field is optional.

**Minimum**: a `(follow-up: decision-attempt-in-engine)` marker here and on the `waits` map, plus the matching `Code marker` line on the rerun ticket, so the rerun work can't miss it.

apps/backend/README.md
```diff
@@ -35,7 +35,24 @@ A node asks a human for a decision by carrying `data.properties.decisionRequest`
...
+| ------ | --------------------------- | ------------------------------------------------------------------------------------------------------------------- |
+| 400    | `validation_error`          | Body shape                                                                                                          |
+| 400    | `invalid_decision`          | Submission against the request; `details[0].code` is a `SUBMITTED_DECISION_ERRORS` key                              |
+| 404    | `execution_not_found`       |                                                                                                                     |
+| 404    | `node_not_found`            | Not in the run's snapshot                                                                                           |
+| 409    | `execution_not_waiting`     | Terminal or cancelling run, or the engine no longer has it                                                          |
+| 409    | `node_not_waiting`          | No request on the node, never parked, or not waiting now. Final                                                     |
```
dawidaksamski: "Final" is true when the engine says node_not_waiting, since the wait is registered before it's announced. It isn't true for the node_never_parked branch at decision.ts:84, which reads the event table: the node_waiting emit activity can lag or retry while the engine already accepts verdicts. Other channels will code against this table, so I'd rather not overpromise.

Suggest splitting the row: engine answer → "Final"; backend count → "Retry after the wait event lands" (or, once the pending-decision resource exists, "read the resource first"). And a clause in decision-log 18: "final when the engine says so; the event-table pre-check can lag the engine."

dawidaksamski (approved):

apps/backend/src/routes/decision.ts
```diff
@@ -0,0 +1,121 @@
...
+]);
+
+const decisionBodySchema = submittedDecisionSchema.extend({
+  nodeId: z.string().min(1),
+  attempt: z.int().min(1),
+});
+
+export function createDecisionRoutes(
+  assertAuthorized: AssertAuthorized,
+): Hono<{ Variables: AuthVariables & TenantVariables }> {
```
librowski: consider extracting a type as it's immediately repeated

apps/backend/src/routes/decision.ts
```diff
@@ -0,0 +1,121 @@
...
+  const routes = new Hono<{ Variables: AuthVariables & TenantVariables }>();
+
+  routes.post('/:id/decision', async (c) => {
+    const executionId = c.req.param('id');
```
librowski, line 42: can we change `c` to `ctx` or `context` unless it's a Hono convention?

apps/backend/src/routes/decision.ts
```diff
@@ -0,0 +1,121 @@
...
+import { logger as backendLogger } from '../logger';
+import type { TenantVariables } from '../tenant';
+import { ENGINE_REFUSALS, LOOKUP_REFUSALS, refuse } from './decision-refusals';
+import { formatValidationDetails } from './snapshot-validation';
+
+const logger = backendLogger.child({ component: 'decision-route' });
+
+// Every other status goes to the engine: the advisory status write is best-effort, so a
+// parked run can still read 'pending'.
+const NOT_DECIDABLE_STATUSES = new Set<string>([
```
librowski, line 28: can we narrow the generic type to make it a `Set<ExecutionStatus>`?

apps/backend/src/routes/decision-refusals.ts
```diff
@@ -0,0 +1,83 @@
...
+export const ENGINE_REFUSALS = {
+  node_not_waiting: 'node_not_waiting',
+  verdict_already_delivered: 'decision_already_made',
+  run_not_found: 'run_gone',
+  verdict_for_unknown_node: 'fault',
+  verdict_malformed: 'fault',
+  delivery_timeout: 'delivery_timeout',
+} as const satisfies Record<ResolveNodeRejection, DecisionRefusal | 'fault'>;
+
+export function refuse(c: Context, refusal: DecisionRefusal, value?: string, extra: Record<string, unknown> = {}) {
```
librowski: I'd consider using an options object parameter for this function.
It's not obvious what `undefined` is when passed as a third parameter for instance.

apps/backend/src/routes/decision.ts
piotrblaszczyk: It is a Hono convention: `c` is what Hono's own docs and types use for the Context. The repo follows it without exception, `async (c)` in all 11 route handlers across `executions`, `workflows`, `visualize` and `decision`, with no `ctx` or `context` anywhere.

Renaming here alone would make this the one file that differs, so I would keep `c`.

apps/backend/src/routes/decision.ts
piotrblaszczyk: `execution.status` is a plain `string`: the column is `text('status')`, not a pg enum, so `Set<ExecutionStatus>.has(execution.status)` would not typecheck. Narrowing the generic would force a cast at the call, and that cast would be a lie, because a text column really can hold anything.

The check you are after is already there. `TERMINAL_EXECUTION_STATUSES` is `as const`, and the one hand-written literal carries `'cancelling' satisfies ExecutionStatus`, so a typo fails the build today. The `string` generic only widens `.has()`. `executions.ts:24` has the same shape for the same reason.

apps/backend/src/routes/decision.ts
piotrblaszczyk: Took the Minimum, deliberately rather than by omission. It is on the branch in `abc8a0fb`: a comment at this check and another on the `waits` map in `run-workflow.ts`, decision-log 15 rewritten to name the non-atomicity instead of implying nothing is owed, and a `Code marker: decision-attempt-in-engine` line on WB-506 carrying your Preferred option written out as the shape to build.

Two reasons for not building it now. Nothing can increment the counter past 1 today, since a node is scheduled at most once per run, so `verdict_attempt_mismatch` would be unreachable. The package README documents every rejection `resolveNode` can answer, so we would be shipping a documented code that cannot fire, in a package queued for its first npm release.

The other reason is that the wait instance's identity is not settled. WB-504's amendment calls it "attempt / waitInstanceId" and that resource is the consumer that decides between a counter and an opaque id. Putting a counter into the update input first picks the less flexible of the two before its consumer exists. Both points are on the ticket, so the rerun work starts from your shape rather than rediscovering the gap.

apps/backend/src/routes/decision.ts
piotrblaszczyk: Done in `c752f77d`. Went a bit wider than this file, because the same type was spelled out nine times: twice in each of the four route factories, once in `server.ts`, and once in the `executions` and `workflows` test harnesses. Fixing only this file would have left eight copies and made `decision.ts` the odd one out.

The alias names the Hono env rather than the whole app type, since `new Hono<...>()` needs the generic either way, so aliasing `Hono<BackendEnv>` would still have been written twice:

```ts
export type BackendEnv = { Variables: AuthVariables & TenantVariables };
```

`AuthVariables` and `TenantVariables` imports dropped out of seven files as a result. No import cycle: neither `auth` nor `tenant` reaches into `routes`.

apps/backend/src/routes/decision-refusals.ts
piotrblaszczyk: Agreed, done in `caf66e00`. Three of the nine call sites passed a bare `undefined` to skip the interpolation value and reach the extra response fields behind it. Both are now one options object:

```ts
refuse(c, 'execution_not_found');
refuse(c, 'node_never_parked', { value: nodeId });
refuse(c, 'attempt_mismatch', { extra: { attempt: waits } });
refuse(c, 'body_invalid', { extra: { details: formatValidationDetails(parsedBody.error) } });
```

`value` fills the message's one `{value}` slot, `extra` adds fields beside `code` and `message`. Every `undefined` is gone and the refusals that need neither still pass nothing.

Checked the tests actually hold this rather than just execute it: ignoring `value` turns 5 tests red, dropping `extra` 14, and passing a value as `extra` at a call site 1.

PR #136, opened 2026-09-14
[DS 2.0] Apply the design body spacing to node sections and rows

librowski:
## Change

Design confirmed the node body matrix (DR-143 answer, 9.09): section padding and gap 8px, row padding 8px, row radius 4px, section radius 8px. Code used provisional 12/10/6 from August.

- `node-section.module.css` and the AI Agent `node-wrapper-info.module.css`: padding and gap → `space/100`, radius → `canvas/node/content-radius` (existing role).
- `connectable-item.module.css`: row padding → `space/100`, radius → `radius/50`.
- The `ConnectableItem` inset follows automatically, so the derived row width is now `241 - 2*(8+1) - 2*(8+1) = 205px` (was 201 with the 10px section padding). Decision log updated with a dated note.

Row padding, row radius, section padding and gap bind to primitives with `/* missing token */` markers: the export has no `node.body.*` roles yet. Design is asked to add `node.body.padding`, `node.body.gap` and row roles; the repoint is a two-line follow-up after the next export.

## Verification

sdk stylelint and unit tests. Browser measurement of the rows (width 205, padding 8, radius 4) is part of the stack's end-to-end pass; see the recipe in the research repo.

Changeset: `@workflowbuilder/sdk` patch.


packages/sdk/src/features/diagram/nodes/components/connectable-item/connectable-item-width.decision-log.md
```diff
@@ -67,3 +67,11 @@ adds horizontal padding without declaring it lets its rows exceed the visible wi
 - Provisional until the design specifies the row width (follow-up: connectable-item-design-width).
   The design's node body matrix (row padding 8px, radius 4px) is a separate change and does not
   alter this derivation.
+
+## Update 14.09.2026
+
+Design confirmed the body matrix (DR-143): section padding and gap 8px, row padding 8px, row radius 4px,
```
piotrblaszczyk, line 73: `DR-143` points at the design-review register outside the repo, the same kind of pointer we took out of this file in 6cfe59db. The repo is public, so for external readers this is a dead reference. Could you drop it and state the fact plainly?

While here: the Decision section above still shows `10 + 1` / `201px` / `0.625rem`, so the log now says two things. I'd fold the new numbers in place and turn `## Update 14.09.2026` into the `## Status` section the other logs use (revision date in the `### Date:` line). A `(follow-up: node-body-roles)` slug on the pending `node.body.*` roles would make the repoint grep-able, like the width follow-up already is.

piotrblaszczyk (approved):

packages/sdk/src/features/diagram/nodes/components/connectable-item/connectable-item-width.decision-log.md
librowski: Rewritten. The Decision section carries the current derivation (inset 8 + 1 per side, 205px, sections and rows on the `body-*` and `row-*` roles from the export), the external register reference is gone and the fact is stated plainly (design confirmed the body matrix on 09.09), and the dated update sections collapsed into a `## Status` block. The only open pointer left is `(follow-up: connectable-item-design-width)`. Fixed in #150.

PR #139, opened 2026-09-14
[DS 2.0] Mark the current choice in Menu and move selected list entries to ui/bg/selected

librowski:
## Change

Design answer to the "selected in lists" questions (DR-029/030/036): a new `ui/bg/selected` role (brand fill subtle, same in both themes) with `selected-hover`, menu items with Selected / Selected+Hover states, selects sharing the same item styling. The role landed in the 14.09 export (#138, this PR's base).

- `list-item.module.css` (shared by `MenuItem` and `SelectOption`): selected background → `ui/bg/selected`, hovered or highlighted selected → `ui/bg/selected-hover`, text → `ui/text/default` instead of on-accent white. The dark-theme override of the old solid `acc1-500` fill goes away, the token carries the theme. New public property `--wb-public-list-item-background-color-selected-hover`.
- `Menu`: items accept `selected`. When any item defines it, the popup renders a Base UI `RadioGroup` and the items become `RadioItem`s (`menuitemradio`, `aria-checked`, `data-checked`, `closeOnClick`). Menus without a selection are unchanged.
- `Select` options already expose `data-selected`, so they pick up the new colors without code changes; `SelectOption` stays because both components already share one stylesheet.
- SDK: the language menu marks the current language.
- Docs: Menu page describes `selected` with an example; the props table picks up the JSDoc.

## Verification

ui typecheck, eslint, stylelint, `menu.spec.tsx` (radio group, `aria-checked`, `data-checked`), build with `check:built-css`; sdk typecheck and language selector spec (3 tests); docs typecheck. Changesets: ui minor, sdk patch.


packages/ui/src/components/menu/menu.tsx
```diff
@@ -77,12 +86,10 @@ export const Menu = memo(
...
           >
             <MenuBase.Popup className={listBoxStyles['list-box']}>
-              {items.map((item, index) =>
-                item.type === 'separator' ? (
-                  <Separator key={index} />
-                ) : (
-                  <MenuItem key={item.label} {...item} size={size} />
-                ),
+              {hasSelection ? (
+                <MenuBase.RadioGroup value={selectedValue}>{renderedItems}</MenuBase.RadioGroup>
```
piotrblaszczyk, line 90: The `RadioGroup` renders a `div[role="group"]` between the popup and the items, so the `.list-box` flex `gap` no longer applies between rows. Measured with the built CSS: a plain menu has 4/4/4 px between items, a radio menu 0/0/0, so the hover and selected fills touch. The SDK language menu ships this.

Could we give the group a class that repeats the column layout, e.g. in `list-box.module.css`: `.group { display: flex; flex-direction: column; align-items: center; align-self: stretch; gap: var(--wb-public-list-box-gap); }` and pass it as `className` to `MenuBase.RadioGroup`? `display: contents` also works, but Safari has dropped the `group` role from the a11y tree with it.

packages/ui/src/components/menu/menu.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+  document.body.append(container);
+  containers.push(container);
+  const root = createRoot(container);
+  act(() => {
+    root.render(<Menu open items={items} />);
+  });
+  return root;
+}
+
+afterEach(() => {
```
piotrblaszczyk, line 20: The roots are never unmounted, and the menu portals into `document.body`, so the items survive `afterEach`: after this file's two tests there are 4 stale `menuitem*` nodes in the document. The assertions pass only because each test queries a different role. Could we mirror the other ui specs: set `IS_REACT_ACT_ENVIRONMENT = true`, keep the `Root`, and `act(() => root.unmount())` in `afterEach`? That also removes the act() warnings this file prints on every run.

piotrblaszczyk (approved):

packages/ui/src/components/menu/menu.tsx
librowski: Good catch. The `RadioGroup` wrapper now gets `.group` from `list-box.module.css` (column flex with `--wb-public-list-box-gap`), so the rows keep the same 4px as a plain menu. Measured in Demo on the language menu: 4px between the radio items, `aria-checked` still correct. Fixed in #150.

packages/ui/src/components/menu/menu.spec.tsx
librowski: Both `menu.spec.tsx` and `node-panel.spec.tsx` now follow the `node-description.spec.tsx` pattern: `IS_REACT_ACT_ENVIRONMENT`, a `Root` per test, `act(() => root.unmount())` and container removal in `afterEach`. Fixed in #150.

PR #146, opened 2026-09-14
feat(examples): runnable Temporal Code Exchange sample

dawidaksamski:
## What

A self-contained sample for [Temporal's Code Exchange](https://temporal.io/code-exchange)
that demonstrates `@workflowbuilder/temporal`: a diagram drawn in the Workflow Builder
editor runs as a durable Temporal Workflow Execution, every node becomes an Activity, and
the canvas shows each node's status live. It lives at `examples/workflow-builder-temporal/`
alongside the starter, outside the pnpm workspace, so `degit` fetches a folder that stands
on its own.

Everything is new except a two-line pointer in the root README.

## What's in it

Three folders, each with one job:

- `shared/` — what both processes agree on: the diagram, the wire types, the coin flip.
- `worker/` — the Temporal Worker with the plugin, the HTTP/SSE bridge the editor talks to,
  and a CLI client. Its own npm project.
- `editor/` — the optional browser app. Its own npm project.

The diagram is five nodes: a trigger, an action that fails on its first attempt on purpose,
a decision, and one action per branch. Two things a reader can see and not just read about:
Event History shows a real retry on Send email, and the branch the runner pruned schedules
no Activity at all — on the canvas it dims out and reports `node_skipped`. The amount is
drawn at random per run, so roughly every other run takes the other branch.

Six places carry the whole integration and the README names each one: the workflow
re-export, the plugin handed to `Worker.create`, the executors, branch evaluation, the
store port, and the snapshot-to-definition mapper.

## Why this is a draft

`@workflowbuilder/temporal` is not on npm yet, so `worker/` installs it from a packed build
committed under `worker/vendor/` and the manifest points at that file. The tarball is a
build of source already public in this repo, so it discloses nothing new, but it is a
stopgap and the sample cannot be submitted to Code Exchange until the package ships.

On publish day (see `worker/vendor/README.md`, code marker
`temporal-sample-npm-dependency`):

1. Delete `worker/vendor/`.
2. Depend on the published range instead of the file.
3. Regenerate `worker/package-lock.json`.
4. Drop the `!/worker/vendor/*.tgz` exception from the sample's `.gitignore`.

Then re-run the clean-machine check against npm — that run is the one that counts, since
the README's `degit` one-liner only resolves once this is on `main`.

## CI

New `temporal-sample` job in `pr-check.yml`, modelled on the existing Starter build: `npm ci`
plus typecheck for the worker, `npm ci` plus build for the editor. No pnpm and no workspace
install, since neither project is a workspace member. It also guards the vendored tarball —
a refreshed build that breaks the sample fails here.

Not covered: ESLint over the sample. Prettier reaches it through the global format job and
the sample's lint-staged config catches it on commit, but a hook-bypassing push would not be
linted. Happy to add a step if reviewers want it.

## How it was verified

- Exported the committed tree with `git archive` into a clean directory — exactly what this
  PR carries, nothing untracked — then `npm ci` in both projects, worker typecheck, editor
  build, and a full run end to end. The sample is self-sufficient.
- Many runs against a local Temporal, forced and random, both branches. Event History
  confirms two attempts on Send email only, and no Activity for the pruned node.
- Editor: Run, live markers, the dimmed skipped node, the Temporal UI link, and Reset
  diagram restoring the shipped diagram.

## Deliberately not here

- No unit tests. The sample demonstrates the plugin rather than testing it; the library's
  own suite lives in `packages/temporal`.
- No docs-site guide — a separate deliverable (`temporal-docs-guide`).
- No Code Exchange submission yet (`temporal-sample-submission`); the README carries the
  text to paste and the pitch fits the form's 256-character limit.
- A human-approval node showing durable pause waits for human-in-the-loop to land in a
  published version (`temporal-sample-approval-node`).

## For reviewers

The README is the deliverable — if the instructions don't work, the sample doesn't work, so
that's the thing to read critically. The two screenshots, the expected output blocks, and
every relative link were taken from or checked against real runs.


examples/workflow-builder-temporal/README.md
```diff
@@ -0,0 +1,174 @@
+# Workflow Builder on Temporal
+
+Run a workflow drawn in the Workflow Builder editor as a durable Temporal Workflow Execution: every node becomes an Activity, retries and history come from Temporal, and the canvas shows each node's status live.
+
+This sample wires the [`@workflowbuilder/temporal`](https://www.npmjs.com/package/@workflowbuilder/temporal) plugin into a Temporal Worker and executes a five-node diagram authored in the [Workflow Builder](https://www.workflowbuilder.io) editor. One node fails on its first attempt on purpose, so the retry you see in Event History is a real one, and a decision node routes each run down one of two branches while the other is pruned. An optional editor lets you change the diagram and watch nodes light up while Temporal runs them.
```
piotrblaszczyk: This links to an npm page that does not exist yet, and line 41 says so. A reviewer will click it in the first paragraph. Until publish day, point it at https://github.com/synergycodes/workflowbuilder/tree/main/packages/temporal, and add "swap the two npm links (README lines 5 and 168)" to the publish-day list in worker/vendor/README.md.

examples/workflow-builder-temporal/README.md
```diff
@@ -0,0 +1,174 @@
...
+- **Tune retries per node type.** Replace the re-export in `worker/src/workflows.ts` with `createRunWorkflow({ nodeActivityProfiles })` and pass the same map to the plugin. The plugin README explains why both sides need it.
+- **Point at Temporal Cloud.** `TEMPORAL_ADDRESS`, `TEMPORAL_NAMESPACE`, and `TEMPORAL_UI_ORIGIN` are read in `worker/src/config.ts`. TLS and API keys go on the two connections in `worker/src/worker.ts` and `worker/src/client.ts`.
+- **Toggle the outage.** Every Action has a _Simulate an outage on the first attempt_ switch. Turn it on for `Auto-approve` and that node retries too.
+- **Add a branch.** Select **Needs review?**, add a branch on the node or in the properties panel, give it a title and conditions, then drag its new handle to a node. Keep the branch with no conditions last, or it swallows every run.
+- **Persist events.** Replace `createRunStore()` with anything that implements `ExecutionStore`.
+
+## Learn more
+
+- [`@workflowbuilder/temporal` on npm](https://www.npmjs.com/package/@workflowbuilder/temporal): profiles, node labels in Event History, versioning and replay.
```
piotrblaszczyk: Same dead link. The plugin README in the repo covers profiles, labels and replay today, so link there for now.

examples/workflow-builder-temporal/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+```bash
+npx degit synergycodes/workflowbuilder/examples/workflow-builder-temporal workflow-builder-temporal
+cd workflow-builder-temporal
+```
+
+Terminal 1, a local Temporal server:
+
+```bash
+temporal server start-dev
```
piotrblaszczyk, line 57: Anyone who already runs Temporal locally has probably 7233 taken. The fallback belongs right here, since `config.ts` already reads the variables:

```bash
temporal server start-dev --port 7234 --ui-port 8234
```

and then `TEMPORAL_ADDRESS=127.0.0.1:7234 TEMPORAL_UI_ORIGIN=http://localhost:8234` in front of `npm start` and `npm run workflow`.

PR #149, opened 2026-09-15
[DS 2.0] Adopt the 15.09 and 16.09 token exports: node body and row roles, focus ring rename, ghost label roles, rebuilt orange

librowski:
## Export

Tokens Studio export from the design system file after the 15.09 publish (changelog entries 1.1.9, 1.1.10, 1.1.11), compared with the 14.09 export by name and value:

| | Count |
|---|---|
| Tokens | 789 → 805 |
| Added | 31 |
| Removed | 15 |
| Changed values | 40 (all in the warning / orange family) |

Added: `canvas/node/body-{h-pad,v-pad,gap,radius}` and `canvas/node/row-{h-pad,v-pad,gap,radius}`; `shadow/canvas/focus-ring/node/{x,y,blur,spread}`; `components/button/ghost/{primary,critical,success,warning}/text-default` and `components/button/solid/warning/text`; `colors/orange-500-{10..50}`.
Removed: `shadow/canvas/focus-node/*` (renamed), `colors/orange-400-{10..50}` (replaced), `canvas/node/inset` and `gap-inner` (orphans confirmed by design).
Changed: the whole `colors/orange-*` scale (rebuilt in OKLCH around #FDAC1B) and every warning alias that follows it.

## Code

- Focus ring: `node-panel.module.css` and AI Studio `highlighting.css` use `--wb-ds-shadow-canvas-focus-ring-node-*`.
- Ghost labels: `--wb-public-button-ghost-{primary,critical,success}-color` bind to the new component roles (light: acc1-700, red-600, green-600; dark unchanged). The shared `ui/text/*-default` roles are untouched, so field labels, destructive menu items, the Select button and AI Studio statuses keep their colours.
- Node body and rows (follow-up of #136): `node-section`, AI Agent `node-info-wrapper` and `connectable-item` bind padding, gap and radius to the body and row roles; the row gap gets its own variable; the last `missing token` markers in the node body are gone. Values unchanged, rows stay 205px. Decision log updated (`connectable-item-design-width` follow-up closed by 1.1.9).
- Nothing in the repo referenced the removed orange tints or the orphan node roles.

Changesets: `@workflowbuilder/ui` minor with a `Breaking changes` note for the renamed ring variables and the replaced orange tints; `@workflowbuilder/sdk` patch.

## Verification

[...]

PR #140, opened 2026-09-14
[DS 2.0] Render palette entries with the canvas Node states

librowski:
## Decision

Design answer to the palette questions (DR-180/181): the palette gets no component and no tokens of its own. An entry is the `Node` without ports and badge; its states are the Node states (rest `Default`, hover `Hover`, dragging `Active` with the same ring as selection, unavailable type `Disabled`). The drag preview is the same Node instance.

## What was already true

`NodePreviewContainer` already renders the real node template with `showHandles={false}`, and the shell's own hover (`stroke-hover`) already applied to palette entries. The palette's `variables.css` (two `--wb-sdk-palette-item-*` properties) had zero consumers and the item wrapper's transparent outline did nothing.

## Change

- `NodePanel.Root`, `NodeDescription`, `NodeIcon` (ui): new `disabled` prop rendering the Node Disabled variant from the Figma master bindings: shell and icon container `canvas/node/bg-primary-disabled`, title and subtitle `canvas/node/text-disabled`, glyph `ui/icon/disabled`, no hover reaction. Each component owns its defaults in `:root` (`--wb-public-node-*-disabled`), per the built-CSS contract that forbids redefining public variables on component selectors.
- Node templates (sdk): `disabled` prop passed to shell, icon and description.
- Palette: read-only entries render `disabled` instead of `opacity: 0.5`; the drag preview renders `selected` (Active ring); dead `variables.css` and outline styling removed.
- Docs: NodePanel, NodeDescription and NodeIcon pages describe the state. Changesets: ui minor, sdk minor.

Measured in the demo on the built package: rest border `gray` / title `#070708` / icon `acc1-500`; hover border `acc1-500`; read-only background `gray-300`, title and icon at 30% alpha, `disabled` class on the shell. Screenshots in the research repo (`visual-pass-2026-09-14/`).

## Follow-up, not in this PR

The container mapping from the same answer (`PaletteContainer` → `Left_Sidebar`): the shared `Sidebar` component is 320px wide with a 20px content gap, the Figma master is 273px with 16px. It also serves the properties bar, so it is a separate change. The drag preview ring is covered by the preview spec, not by a browser check (native drag image).

## Verification

ui and sdk typecheck, eslint, stylelint, unit tests (NodePanel states, NodeDescription disabled, preview state passthrough), `pnpm -F @workflowbuilder/ui build` with `check:built-css`, docs typecheck.

PR #150, opened 2026-09-15
fix: address review follow-ups on the canvas stack

librowski:
## Summary

Follow-ups from the second review round on the canvas stack (#124, #126, #136, #139, #140), plus the code side of the design answers from 16.09. The reviewed PRs are approved, so the fixes land here on top of #149.

## Review follow-ups (commit 1)

**Persist without measured (#124)**
- `getNodesWithoutDynamicValues` also strips the runtime `dragging` flag; the bundled Demo and AI Studio templates no longer carry `dragging: false`.
- `getStoreDataForIntegration` gets two boundary tests; changeset mentions `dragging`.

**Node text ellipsis (#126)**
- Docs and changeset state the `contain: inline-size` constraint (definite inline size on an ancestor; collapses inside `inline-flex` or `fit-content`).

**Node body spacing (#136)**
- Width decision log rewritten with the current derivation and a `## Status` section, no external tracker reference; tooltip changeset reworded.

**Menu item selected (#139)**
- `Menu.RadioGroup` wrapper gets `.group` (column flex, `--wb-public-list-box-gap`), measured 4px between radio items; `menu.spec.tsx` and `node-panel.spec.tsx` unmount the React root after each test.

**Palette node states (#140)**
- The custom node guide and the Demo multi-port template forward `disabled` to `NodePanel.Root`, `NodeIcon` and `NodeDescription`; `node-panel.mdx` lists `disabled?: boolean`; changeset written around `WorkflowNodeTemplateProps.disabled` with a migration note.

## Design answers of 16.09 (commit 2)

- **Rows fill the section.** In the horizontal layout Decision branches and AI tools take the section width like the design's `Node / Row` (`width: 100%`, `min-width: 0`, no cap); 205px at 241 is the reference value (design draws 209 because a Figma stroke takes no layout space). The derived cap and the Decision `min-width: max-content` rule now apply to the DOWN layout only, which has no design master. Branch rows and tool rows are spaced with `--wb-ds-canvas-node-body-gap` (8px; was 4px hard-coded in both places). Decision log updated, the `connectable-item-design-width` follow-up is closed.
- **Self-loop apex at 48px above the node's top edge**, independent of node height and port position. `SELF_CONNECTING_EDGE_LABEL_OFFSET` is `48` and measured from the top edge, as its JSDoc always said; `getSelfLoopHeight` is exported. Measured in Demo: 48px in the RIGHT layout; 44px in DOWN, where xyflow anchors the bottom port on its outer edge.
- **DS Tooltip decided for 3.1**: `(follow-up: node-text-ds-tooltip)` markers on the `title` attributes in `NodeDescription` and `ConnectableItem`.
- Demo multi-port template forwards `disabled` to `NodeIcon` and `NodeDescription` (review remark on commit 1).

## Checks

- `pnpm -F @workflowbuilder/ui build` (includes `check:built-css`), vitest ui 74 and sdk 219 passing, stylelint ui/sdk clean, typecheck ui/sdk/demo/ai-studio clean.
- Demo measurements (seed with long labels, RIGHT): Decision 241px, rows 205px with truncation, row gap 8px, tool gap 8px, loop apex 48px. DOWN unchanged (525px, capped row 205px).


apps/demo/src/app/components/multi-port-node/multi-port-node-template.tsx
```diff
@@ -40,7 +41,7 @@ export const MultiPortNodeTemplate = defineNodeTemplate<MultiPortProperties>(
       return (
         <div className={styles['wrapper']}>
           <div className={barClassName} />
-          <NodePanel.Root selected={selected}>
+          <NodePanel.Root selected={selected} disabled={disabled}>
             <NodePanel.Header>
               <NodeIcon icon={iconElement} />
```
piotrblaszczyk: The guide snippet and the changeset's migration note forward `disabled` to all three parts, but here only `NodePanel.Root` gets it. `NodeIcon` and `NodeDescription` mute only through their own prop, and the shell's `disabled` swaps just the background, so the read-only palette shows this entry with a faded surface but full-contrast icon and text. Could we pass `disabled={disabled}` on lines 46-47 as well?

piotrblaszczyk (approved):

apps/demo/src/app/components/multi-port-node/multi-port-node-template.tsx
librowski: NodeIcon and NodeDescription now get `disabled={disabled} `too, so the Demo entry matches the guide and the
  migration note. Fixed in `f925eae1`, which also carries the code side of the design answers from 16.09 (rows fill, loop offset 48px, `Tooltip` follow-up markers).

PR #153, opened 2026-09-15
chore: type-check spec files in demo and ai-studio

librowski:
## Summary

`apps/demo` and `apps/ai-studio` excluded `src/**/*.spec.*` and `src/**/*.test.*` from their tsconfig, so the pre-commit `tsc --noEmit` from lint-staged and `pnpm check` never type-checked tests and stayed green with type errors in them. CI does not type-check these apps by design (`pr-check.yml` header). Both exclusions come from the initial 2.0.0 import, before either app had a test. `packages/sdk` and `packages/ui` already type-check their specs.

Neither tsconfig emits anything (vite builds the apps), so the exclusion only removed a check.

## Changes

- `apps/demo/tsconfig.json` and `apps/ai-studio/tsconfig.json`: `exclude` keeps only `dist`; `"vitest/globals"` joins `types`, as in the packages, so a spec written against `test.globals` (no vitest import) type-checks instead of failing with TS2593.

## Checks

- `pnpm -F @workflow-builder/demo typecheck` and `pnpm -F ai-studio typecheck` pass with no code changes.
- A deliberate `const probe: number = 'x'` appended to `detect-format.test.ts` fails the AI Studio typecheck; a spec using bare `describe`/`it`/`expect` type-checks with `vitest/globals` and fails without it.
- AI Studio tests: 16 passing.


apps/ai-studio/tsconfig.json
```diff
@@ -9,5 +9,5 @@
     "verbatimModuleSyntax": true,
     "types": ["node", "vite/client", "vite-plugin-svgr/client"]
```
piotrblaszczyk: Since `test.globals: true` is on in both apps, could you add `"vitest/globals"` to `types` here and in `apps/demo/tsconfig.json`, like `packages/sdk` and `packages/ui` do? Otherwise a spec written in the ui style (`describe` without an import) passes vitest but fails the new typecheck with `TS2593` and a confusing `@types/jest` hint. I checked: with it added, such a spec type-checks and the app stays green.

apps/ai-studio/tsconfig.json
```diff
@@ -9,5 +9,5 @@
...
   },
-  "exclude": ["src/**/*.spec.ts", "src/**/*.test.ts", "src/**/*.spec.tsx", "src/**/*.test.tsx", "dist"]
+  "exclude": ["dist"]
```
piotrblaszczyk, line 12: Small accuracy note on the summary and the commit message: `pr-check.yml` never type-checks demo or ai-studio (its header comment says so), so CI was not the thing staying green. The guard this restores is the pre-commit `tsc --noEmit` from lint-staged and `pnpm check`. Worth rewording before the squash so `git blame` tells the right story.

piotrblaszczyk (approved):

apps/ai-studio/tsconfig.json
librowski: Added to `types` in both apps. Reproduced your case first: a spec with bare `describe`/`it`/`expect` passed vitest and failed the typecheck with TS2593 and the `@types/jest` hint; with the type it is clean. Fixed in the amended commit.

apps/ai-studio/tsconfig.json
librowski: You are right, `pr-check.yml` never covered these apps. Commit message and PR body now say that the guard being restored is the pre-commit `tsc --noEmit` from lint-staged and `pnpm check`, and that CI leaves demo and ai-studio out by design. Whether CI should start covering them is a separate conversation.

